Issue #10, opened 2018-11-29
Create a test runner for small code snippets


hermanventer:
We can expect to accumulate a large collection of small program snippets, each of which exercises some small part of Mirai. It is not convenient to have to write a new entry in tests/integration_tests.rs for each snippet and to have to recompile the ever growing tests/integration_tests.rs every time we add a snippet.

So, instead, let's write a dedicated test runner that will invoke mirai independently on each snippet found in a dedicated directory (e.g. tests/snippets). There should be a mode that can run a designated snippet for debugging purposes (BTW. I can't figure out how to debug a test run via cargo test with VSCode. Please comment if you know how.)

Preferably a default run should use all of the cores of the machine it is running on and should produce minimal console spew for successful runs. When one or more tests fail, the console output should be no more than is needed to run any of the failing tests as an individual run inside a debugger.

fpoli:
An interesting crate to look at for this is compiletest-rs, which we also use in Prusti. It can be used, for example, to test that the compiler driver emits determined error messages.

(Sorry for not providing more links, I'm from the mobile)

hermanventer:
This sounds like just the thing to use.

hermanventer:
I spent some more time looking into the possibility of leveraging compiletest-rs and the underlying logic in the rustc compiler, while not creating a new process for every test run. So far, I have not come across a simple way to reuse code from these sources, partly because so much of it is not public.

PR #14, opened 2018-12-07
Stubs for visiting MIR nodes

### LINKED ISSUE #11: Set up visitor infrastructure for MIR
Write the code to 
a) find every function/method body in the crate being compiled and
b) load MIR for the body and then traverse the body, visiting every MIR instruction for abstract interpretation.

hermanventer:
## Description

Flesh out the visitor class to include stubs for MIR nodes.

Fixes #11

## Type of change

## How Has This Been Tested?

tests/run-pass/crate_traversal.rs has been fleshed out with bodies that cause all the new visitor methods to get called.

### DIFF
--- a/tests/run-pass/crate_traversal.rs
+++ b/tests/run-pass/crate_traversal.rs
@@ -1,16 +1,164 @@
-// A small trait to test crate traversal
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
 
-trait TestTrait {
-    fn test_method() {}
+// statements and expressions that exercise all visitor code paths
+
+#![allow(const_err)]
+#![allow(unused)]
+#![feature(asm)]
+#![feature(box_syntax)]
+#![feature(generators)]
+
+use std::mem;
+
+pub trait TestTrait {
+    fn test_method() {
+        let _loc = [1];
+        let loc2 = [ 0; 512 ];
+    }
 }
 
-struct TestStruct {
-     test_field: i64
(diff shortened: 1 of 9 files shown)


tests/run-pass/crate_traversal.rs
```diff
@@ -1,16 +1,164 @@
...
 }
 
 impl TestStruct {
-    fn test_function() {}
-    fn test_method(&self) {}
+    pub fn test_function() {}
+    pub fn test_method(&self) {}
+}
+
+pub unsafe fn test1() {
```
sblackshear: Some test method names and/r comments describing which features each test is targeting might be useful? I can figure some of them out, but not all :).

tests/run-pass/crate_traversal.rs
```diff
@@ -1,16 +1,164 @@
...
+}
+
+fn test2() {
+    let z = 0;
+    let _x = {
+        &z
+    };
+}
+
+fn test3() {
```
sblackshear: How do these tests work? Is there an expected output file somewhere?

hermanventer: At the moment the role of the test file is just to trigger calls to all of the visitor methods. They don't do anything right now, so there is no output.

tests/run-pass/crate_traversal.rs
hermanventer: In general, yes, such comments are very helpful (and I wish the rust compiler's test suite had more of them). This test file, however, is of a temporary nature and will become redundant as more useful tests with expected behavior and so on come along. See issue #10. Adding comments detailing with MIR features each test uses would be very tedious and is not a top priority for me right now.

PR #15, opened 2018-12-14
Summaries

### LINKED ISSUE #5: Figure out how summaries will be stored and how to load them incrementally
Mirai computes a summary for each method and function it analyzes. These summaries should be stored persistently so that they can be retrieved during incremental compilation.

Decide on a way to do this and initialize a cache for summaries when processing the input used by the compilation.

hermanventer:
## Description

This provides the outlines of what a function summary looks like and provides a summary cache that is backed by a RocksDB database.

Fixes #5 

## Type of change

## How Has This Been Tested?

Existing tests still cover most of this. Also added a test to exercise the disambiguation logic.


src/abstract_value.rs
```diff
@@ -0,0 +1,114 @@
...
+/// as well, since all operations can be uniform.
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Hash)]
+pub struct AbstractValue {
+    /// An abstract value is the result of some expression.
+    /// The source location of that expression is stored in provence.
+    /// When an expression is stored somewhere and then retrieved via an accessor expression, a new
+    /// abstract value is created (via refinement using the current path condition) with a provence
+    /// that is the source location of accessor expression prepended to the provence of the
+    /// refined expression.
+    pub provenance: List<Span>,
```
sblackshear: Can this be empty?

And as a side note, why `List` instead of `Vec`? I'm guessing that the Rust compiler is more clever about optimizing uses of `Vec` than third-party collections, but I have no evidence of this.

src/abstract_value.rs
```diff
@@ -0,0 +1,114 @@
...
+    StaticRoot {
+        name: Name,
+    },
+
+    /// The identifier denotes a field in the struct found at the path in qualifier.
+    /// For example, in struct A { f: i32 }; fn f(x: A) { x.f }, we identify x.f with
+    /// Path::QualifiedName { qualifier: Path::Parameter { offset: 0 }, selector: String::from("f") }
+    QualifiedName {
+        qualifier: Box<Path>,
+        selector: String,
```
sblackshear: Does the structure of MIR prevent us from having to handle `x.f.g = ...` by using multiple selectors here? That would be great.

src/callbacks.rs
```diff
@@ -140,10 +107,44 @@ impl<'a> CompilerCalls<'a> for MiraiCallbacks {
 /// interpretation of all of the functions that will end up in the compiler output.
 fn after_analysis(state: &mut driver::CompileState) {
     let tcx = state.tcx.unwrap();
+    let crate_name: &str = state.crate_name.unwrap();
+    let summary_store_path: String = String::from(".summary_store.rocksdb");
```
sblackshear: Is there a risk of concurrent tests clobbering each other because this path is always the same?

src/summaries.rs
```diff
@@ -0,0 +1,185 @@
...
+/// This is calculated once per function and is used by callers of the function.
+/// Callers will specialize this summary by replacing embedded parameter values with the corresponding
+/// argument values and then simplifying the resulting values under the current path condition.
+///
+/// Summaries are stored in a persistent, per project database. When a crate is recompiled,
+/// all summaries arising from the crate are recomputed, but the database is only updated when
+/// a function summary changes. When this happens, all callers of the function need to be reanalyzed
+/// using the new summary, which could result in their summaries being updated, and so on.
+/// In the case of recursive loops, a function summary may need to be recomputed and widened
+/// until a fixed point is reached. Since crate dependencies are acyclic, fixed point computation
```
sblackshear: > crate dependencies are acyclic

This is great; it simplifies things so much for us compared to the Z/Infer setting!

src/summaries.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub preconditions: List<AbstractValue>,
+
+    // If the function returns a value, this summarizes what is known statically of the return value.
+    // Callers should substitute parameter values with argument values and simplify the result
+    // under the current path condition.
+    pub result: Option<AbstractValue>,
+
+    // Modifications the function makes to mutable state external to the function.
+    // Every path will be rooted in a static or in a mutable parameter.
+    // No two paths in this collection will lead to the same place in memory.
```
sblackshear: No two paths or no two *distinct* paths? And if the former, would it make sense to use a map to enforce the  invariant that keys or disjoint (or wrap `List<(Path, AbstractValue)>` in something that enforces this)?

src/summaries.rs
```diff
@@ -0,0 +1,185 @@
...
+    // side-effects of the call.
+    pub side_effects: List<(Path, AbstractValue)>,
+
+    // Conditions that should hold subsequent to the call.
+    // Callers should substitute parameter values with argument values and simplify the results
+    // under the current path condition. The resulting values should be treated as true, so any
+    // value that is not the actual value true, should be added to the current path conditions.
+    pub post_conditions: List<AbstractValue>,
+
+    // Condition that if true imply that the call to the function will not complete normally
```
sblackshear: imply => implies

src/abstract_value.rs
hermanventer: List is a functional data structure, so it enables more sharing than Vec does. Also, we will be prepending to provenance, which is not O(1) for Vec.

src/abstract_value.rs
hermanventer: There will be multiple selectors, but they are hiding away inside the qualifier. I'm a bit miffed at having to own (i.e. duplicate) the nested path, but that is required for deserialization.

src/callbacks.rs
hermanventer: There is, hence the removal of the other tests. As part of dealing with issue #10 I'll address this by adding a parameter that can create a unique temporary location for the DB.

src/summaries.rs
hermanventer: Indeed. :-)

src/summaries.rs
hermanventer: Good point. Let me think about this a bit more.

PR #16, opened 2018-12-15
Update the developer guide and tweak log output

hermanventer:
## Description

Update the developer guide with more information about DYLD_LIBRARY_PATH and tweak log output so that info is a bit more informative.

Since the above kept failing the CI build, also changed the CI script to be more standard (and do more caching) and changed the callback code to use the output directory as the location of the persistent summary cache. Then also changed the integration test script to use a different output directory every time.

# Type of change

## How Has This Been Tested?

Built Rust libstd with MIRAI_LOG=info and observed successful completion of build.

### DIFF
--- a/documentation/DeveloperGuide.md
+++ b/documentation/DeveloperGuide.md
@@ -35,7 +35,7 @@ actually use Clion for editing, VSCode for debugging and Nuclide for doing sourc
 Running `cargo build` will produce a binary at `target/debug/mirai`.
  
 Unfortunately cargo build sets the dynamic library load path (rpath) that is linked into the binary to a path that is
-invalid a when the binary is run. If you run the binary via cargo `cargo run -- <args>` then cargo overrides the bad
+invalid when the binary is run. If you run the binary via cargo `cargo run -- <args>` then cargo overrides the bad
 rpath by providing the correct path in the environment variable `DYLD_LIBRARY_PATH`.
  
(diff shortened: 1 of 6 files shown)


documentation/DeveloperGuide.md
```diff
@@ -32,15 +32,21 @@ actually use Clion for editing, VSCode for debugging and Nuclide for doing sourc
 
 ## Running
 
-Running `cargo build` will produce a binary at `target/debug/mirai`. Create a handy alias for the binary with
+Running `cargo build` will produce a binary at `target/debug/mirai`.
+ 
+Unfortunately cargo build sets the dynamic library load path (rpath) that is linked into the binary to a path that is
+invalid a when the binary is run. If you run the binary via cargo `cargo run -- <args>` then cargo overrides the bad
```
zlandau: "a when" -> "when"

   If you run the binary via cargo `cargo run -- <args>` then cargo overrides the bad

->

   If you run the binary via `cargo run -- <args>` then cargo overrides the bad

documentation/DeveloperGuide.md
```diff
@@ -69,6 +75,12 @@ configurations property of the content of the launch.json file in the .vscode di
     },
 ```
 
+Note that VSCode runs cargo to build mirai (if necessary) and get's the location of the binary from cargo. When
```
zlandau: "get's" -> "gets"

PR #23, opened 2018-12-22
Create abstract values for referenced function constants

hermanventer:
## Description

This sets up some scaffolding for creating a test case that results in a genuine diagnostic, namely that a call to std::intrinsics::unreachable can actually be reached.

The changes in this request are mostly fleshing out of abstract values, but also involves some fixes to caching logic and more functionality in the visitor.

## Type of change

## How Has This Been Tested?

Travis

I had to disable code coverage checking since Tarpaulin crashes with a segfault.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -3,8 +3,38 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use constant_value::ConstantValue;
+
+// See https://github.com/facebookexperimental/MIRAI/blob/master/documentation/AbstractValues.md.
+
+/// A collection of abstract domains associated with a particular abstract value.
+/// The expression domain captures the precise expression that resulted in the abstract
+/// value. It can be used to derive any other kind of abstract domain on demand.
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Hash)]
 pub struct AbstractDomains {
-    //todo: just store the expression (as a domain)
-//todo: use cached getters to get other domains on demand
+    pub expression_domain: ExpressionDomain,
+    //todo: use cached getters to get other domains on demand
 }
+
+/// The things all abstract domain instances have to be able to do
+pub trait AbstractDomain {
+    //todo: join, is_top, is_bottom, etc.
+}
+
+/// An abstract domain uses a functional (side effect free) expression that precisely tracks
+/// a single value.
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Hash)]
+pub enum ExpressionDomain {
+    /// An expression that represents any possible value
+    Top,
+
+    /// An expression that represents an impossible value, such as the value returned by function
+    /// that always panics.
+    Bottom,
+
+    /// An expression that is a compile time constant value, such as a numeric literal or a
+    /// function.
+    CompileTimeConstant(ConstantValue),
+}
+
+// todo: implement the AbstractDomain trait for ExpressionDomain
--- a/src/constant_value.rs
+++ b/src/constant_value.rs
@@ -1,0 +1,99 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+use rustc::hir::def_id::DefId;
+use std::collections::HashMap;
+use summaries::PersistentSummaryCache;
+
+/// Abstracts over constant values referenced in MIR and adds information
+/// that is useful for the abstract interpreter. More importantly, this
+/// value can be serialized to the persistent summary cache.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum ConstantValue {
+    /// A reference to a function
+    Function {
+        /// Indicates if the function is known to be treated specially by the Rust compiler
+        is_intrinsic: bool,
+        /// The key to use when retrieving a summary for the function from the summary cache
+        summary_cache_key: String,
+        // todo: is there some way store the def_id here if available?
+        // this would not be serialized/deserialized.
+    },
+    // A place holder for other kinds of constants. Eventually this goes away.
+    Unimplemented,
+}
+
+impl ConstantValue {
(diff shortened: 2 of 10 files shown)


src/abstract_domains.rs
```diff
@@ -3,8 +3,38 @@
...
+/// An abstract domain uses a functional (side effect free) expression that precisely tracks
+/// a single value.
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Hash)]
+pub enum ExpressionDomain {
+    /// An expression that represents any possible value
+    Top,
+
+    /// An expression that represents an impossible value, such as the value returned by function
+    /// that always panics.
+    Bottom,
```
sblackshear: Probably overkill at this point, but: in Infer we had functors for lifting a value`CompileTimeConstant(ConstantValue)` to an `AbstractDomain` by adding artificial top and bottom elements that are automatically incorporated into the partial order operations. 

I haven't figured out how to do this well in Rust yet--the lack of higher-kinded types means there's no direct analogue of the functor approach. On the other hand, Rust macros might be powerful enough to do this, but tend to make code very hard to read...

src/constant_value.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+use rustc::hir::def_id::DefId;
```
sblackshear: What is this `DefId` thing? Is a `DefId` globally unique?

src/constant_value.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// Does an expensive check to see if the given function is std::intrinsics::unreachable.
+    /// Once it finds the function it caches it so that subsequent checks are cheaper.
+    pub fn check_if_std_intrinsics_unreachable_function(&mut self, fun: &ConstantValue) -> bool {
+        let result = match &self.std_intrinsics_unreachable_function {
+            Some(std_fun) => Some(*std_fun == *fun),
+            _ => None,
+        };
+        match result {
```
sblackshear: Nit: `result.unwrap_or_else(<logic for None case>)` instead of pattern matching?

src/constant_value.rs
```diff
@@ -0,0 +1,99 @@
...
+            None => {
+                let result = match fun {
+                    ConstantValue::Function {
+                        is_intrinsic,
+                        summary_cache_key,
+                    } => *is_intrinsic && summary_cache_key.ends_with("unreachable"),
+                    _ => false,
+                };
+                if result {
+                    self.std_intrinsics_unreachable_function = Some(fun.clone());
```
sblackshear: Nit: `match` on `self.std_intrinsics_unreachable_function` instead to avoid `clone`?

```
match self.std_intrinsics_unreachable_function {
  None => false,
  Some(unreachable_fun) => *unreachable_fun == *fun,
}
```

src/constant_value.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Given the MIR DefId of a function return the unique ConstantValue that corresponds to
+    /// the function identified by that DefId.
+    pub fn get_function_constant_for(
+        &mut self,
+        def_id: DefId,
+        summary_cache: &mut PersistentSummaryCache,
+    ) -> &ConstantValue {
+        self.cache
+            .entry(def_id)
+            .or_insert_with(|| ConstantValue::for_function(def_id, summary_cache))
```
sblackshear: It feels a bit odd to me that we cache this in both `self.cache` and `summary_cache`. Do we need both?

src/constant_value.rs
```diff
@@ -0,0 +1,99 @@
...
+            is_intrinsic: false,
+            summary_cache_key: summary_cache_key.to_owned(),
+        }
+    }
+}
+
+/// Keeps track of MIR constants that have already been mapped onto ConstantValue instances.
+pub struct ConstantValueCache {
+    cache: HashMap<DefId, ConstantValue>,
+    std_intrinsics_unreachable_function: Option<ConstantValue>,
```
sblackshear: Will we need a new field for each intrinsic function we care about?

src/abstract_domains.rs
```diff
@@ -3,8 +3,38 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use constant_value::ConstantValue;
+
+// See https://github.com/facebookexperimental/MIRAI/blob/master/documentation/AbstractValues.md.
+
+/// A collection of abstract domains associated with a particular abstract value.
+/// The expression domain captures the precise expression that resulted in the abstract
```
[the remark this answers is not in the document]
hermanventer: That is not precisely what I mean here. :-) But the comment will be better off without this word.

src/abstract_domains.rs
hermanventer: I am very much opposed to macros and not very fond of higher-kinded types. For me, code readability is the only thing that matters and I find such code hard to read.

src/constant_value.rs
hermanventer: It is short for Definition Identifier and it is a handle that you can use to queries in a single compiler session. It is not globally unique. Things may change on this front, apparently.

The information retrieved from queries can be computed on demand, so that the compiler does not hang on to too much memory.

src/constant_value.rs
hermanventer: I don't think so. Most of them can be modeled by summaries. It is possible that unreachable itself can be modeled by a  summary with a false precondition, but it will take quite a bit more infrastructure to be present before that will be actually useful. So, this is probably temporary scaffolding.

src/constant_value.rs
hermanventer: In this instance the summary cache is used to hold on to the summary cache key, which gets computed from the def_id in a not very cheap way.

src/constant_value.rs
hermanventer: That would be nice, but the false result in the None case would be incorrect. The function comment tries to explain this:

    /// Does an expensive check to see if the given function is std::intrinsics::unreachable.
    /// Once it finds the function it caches it so that subsequent checks are cheaper.

PR #25, opened 2019-01-02
Give warning if control might right reach std::intrinsics::unreachable

hermanventer:
## Description

This follows up on PR #23 and actually emits a warning if the visitor comes across a call to std::intrinsics::unreachable.

## Type of change

## How Has This Been Tested?

Existing tests

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,8 @@
 ignore:
   - tests/.*
   - src/main.rs
+
+coverage:
+  range: 60..85
+  round: down
+  precision: 0
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 # install linter
 - rustup component add clippy-preview
 # install code coverage tool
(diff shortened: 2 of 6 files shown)


hermanventer:
Ignore the code coverage results. The code coverage tool is buggy at the moment.

hermanventer:
I am pretty allergic to failing tests myself. The specific failure of the last test run before you looked at this appears to because of a breaking change in the options of Tarpaulin. Hopefully, fixing that will get the run to succeed. I do, however, see many failures that appear to be Heisenbugs. And then the coverage results are incomplete when macros get into the picture, so coverage tests sometimes show up as regressions even when everything else is fine.

If things were consistently broken, I'd disable, but I hate to give up the value of seeing code coverage results for at least some pull requests.

The warnings you see from running Mirai on itself are false positives and this is expected right now since Mirai is far from precise right now.

zlandau:
Ok, changes look fine then.

PR #26, opened 2019-01-04
Let test runner intercept diagnostic messages

### LINKED ISSUE #10: Create a test runner for small code snippets

hermanventer:
## Description

In order to let the test runner check if Mirai produces the expected diagnostic messages for a particular test case it is necessary to provide a way to intercept and collect the diagnostics produced during analysis. This is done by providing the two call back functions:
1) A call back to do the interception and collection
2) A call back that receives the collected messages and then compares it to expected messages extracted from the test case source.

The default handlers for these call backs just emits the diagnostic in case 1 and does nothing in case 2.

Fixes #10 

## Type of change

## How Has This Been Tested?

Modified the unreachable test case to include expected output. Checked that the test runner fails when the expected output is missing or incorrect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_script:
 # install formatter
 - rustup component add rustfmt-preview
 # install linter
-- rustup component add clippy-preview
+#- rustup component add clippy-preview
 # install code coverage tool
 - RUSTFLAGS="--cfg procmacro2_semver_exempt" cargo install --force cargo-tarpaulin || true
 
@@ -25,7 +25,7 @@ script:
 # Run format checks
 - cargo fmt --all -- --check
 # Run lint checks
-- cargo clippy -- -D warnings
+#- cargo clippy -- -D warnings
 # Build
 - cargo build
 # Run unit and integration tests
--- a/src/visitors.rs
+++ b/src/visitors.rs
@@ -12,10 +12,16 @@ use rustc::ty::{Const, Ty, TyCtxt, TyKind, UserTypeAnnotationIndex};
 use rustc::{hir, mir};
 use std::borrow::Borrow;
 use summaries::PersistentSummaryCache;
+use syntax::errors::{Diagnostic, DiagnosticBuilder};
 use syntax_pos;
 
 /// Holds the state for the MIR test visitor.
 pub struct MirVisitor<'a, 'b: 'a, 'tcx: 'b> {
+    /// A place where diagnostic messages can be buffered by the test harness.
+    pub buffered_diagnostics: &'a mut Vec<Diagnostic>,
+    /// A call back that the test harness can use to buffer the diagnostic message.
+    /// By default this just calls emit on the diagnostic.
+    pub emit_diagnostic: fn(&mut DiagnosticBuilder, buf: &mut Vec<Diagnostic>) -> (),
     pub session: &'tcx Session,
     pub tcx: TyCtxt<'b, 'tcx, 'tcx>,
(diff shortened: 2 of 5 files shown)


tests/integration_tests.rs
```diff
@@ -90,11 +97,75 @@ fn invoke_driver(file_name: String, temp_dir_path: String, sys_root: String) {
...
+            panic!("Unexpected error: {} Expected: {:?}", msg, self.messages);
+        }
+    }
+}
+
+/// Scans the contents of test file for patterns of the form "//~ message"
+/// and returns a vector of the matching messages.
+fn load_errors(testfile: &Path) -> Vec<String> {
+    let rdr = BufReader::new(File::open(testfile).unwrap());
+    let tag = "//~".to_string();
```
zlandau: Not to important, but does this really need to be a String?

tests/integration_tests.rs
```diff
@@ -90,11 +97,75 @@ fn invoke_driver(file_name: String, temp_dir_path: String, sys_root: String) {
...
+    rdr.lines()
+        .enumerate()
+        .filter_map(|(_line_num, line)| parse_expected(&line.unwrap(), &tag))
+        .collect()
+}
+
+/// Returns the message part of the pattern "//~ message" if there is a match, otherwise None.
+fn parse_expected(line: &str, tag: &str) -> Option<String> {
+    let start = line.find(tag)? + tag.len();
```
zlandau: TIL you can use ? for Option as well.

src/visitors.rs
```diff
@@ -12,10 +12,16 @@ use rustc::ty::{Const, Ty, TyCtxt, TyKind, UserTypeAnnotationIndex};
 use rustc::{hir, mir};
 use std::borrow::Borrow;
 use summaries::PersistentSummaryCache;
+use syntax::errors::{Diagnostic, DiagnosticBuilder};
 use syntax_pos;
 
 /// Holds the state for the MIR test visitor.
 pub struct MirVisitor<'a, 'b: 'a, 'tcx: 'b> {
+    /// A place where diagnostic messages can be buffered by the test harness.
```
zlandau: If you end up adding more than just these two callbacks you might want to consider a trait instead.

.travis.yml
```diff
@@ -15,7 +15,7 @@ before_script:
 # install formatter
 - rustup component add rustfmt-preview
 # install linter
-- rustup component add clippy-preview
+#- rustup component add clippy-preview
```
zlandau: Were these removals intentional?

hermanventer: Sadly, yes. Nightly currently does not support Clippy.

src/visitors.rs
hermanventer: The thought crossed my mind as well, but hopefully this is it and it seems more light weight.

PR #27, opened 2019-01-08
Add fixed point iteration to visit_body

hermanventer:
## Description

Flesh out the logic in the visit_body method of MirVisitor to compute a fixed point of input states for each basic block before visiting the blocks to look for problems and issuing diagnostics.

Since this involves joining/widening and checking if a state is a subset of another state, a lot of scaffolding has been added to the abstract value/domain classes. Various pieces of existing code have also been tweaked now that things are a bit more concrete.

Also new this PR is a separate class to model the memory (environment) that is accessible to a block.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Added a new test case, but for 100% coverage we'll have to wait for more scaffolding.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -115,10 +115,10 @@ pub enum ExpressionDomain {
     /// An expression that is a compile time constant value, such as a numeric literal or a function.
     CompileTimeConstant(ConstantValue),
 
-    /// An expression that is either if_true or if_false, depending on the value of join_condition.
-    JoinedExpression {
+    /// An expression that is either if_true or if_false, depending on the value of condition.
+    ConditionalExpression {
         // A condition that results in a Boolean value
-        join_condition: Box<AbstractDomains>,
+        condition: Box<AbstractDomains>,
         // The value of this expression if join_condition is true.
         if_true: Box<ExpressionDomain>,
         // The value of this expression if join_condition is false.
@@ -148,20 +148,11 @@ impl AbstractDomain for ExpressionDomain {
     /// In a context where the join condition is known to be true, the result can be refined to be
     /// just self, correspondingly if it is known to be false, the result can be refined to be just other.
     fn join(&self, other: &ExpressionDomain, join_condition: &AbstractDomains) -> ExpressionDomain {
-        ExpressionDomain::JoinedExpression {
-            join_condition: box join_condition.clone(),
+        ExpressionDomain::ConditionalExpression {
+            condition: box join_condition.clone(),
             if_true: box self.clone(),
             if_false: box other.clone(),
         }
-    }
-
-    /// Returns a domain whose corresponding set of concrete values include all of the values
-    /// corresponding to self and other.The set of values may be less precise (more inclusive) than
-    /// the set returned by join. The chief requirement is that a small number of widen calls
-    /// deterministically lead to Top.
-    fn widen(&self, _other: &Self, _join_condition: &AbstractDomains) -> ExpressionDomain {
-        //todo: don't get to top quite this quickly.
-        ExpressionDomain::Top
     }
 
     /// True if all of the concrete values that correspond to self also correspond to other.
@@ -177,7 +168,7 @@ impl AbstractDomain for ExpressionDomain {
             // The universal set is not a subset of any set other than the universal set
             (ExpressionDomain::Top, _) => false,
             (
-                ExpressionDomain::JoinedExpression {
+                ExpressionDomain::ConditionalExpression {
                     if_true, if_false, ..
                 },
                 _,
@@ -187,7 +178,7 @@ impl AbstractDomain for ExpressionDomain {
             }
             (
                 _,
-                ExpressionDomain::JoinedExpression {
+                ExpressionDomain::ConditionalExpression {
                     if_true, if_false, ..
                 },
             ) => {
@@ -200,5 +191,14 @@ impl AbstractDomain for ExpressionDomain {
                 ExpressionDomain::CompileTimeConstant(cv2),
             ) => cv1 == cv2,
         }
+    }
+
+    /// Returns a domain whose corresponding set of concrete values include all of the values
+    /// corresponding to self and other.The set of values may be less precise (more inclusive) than
+    /// the set returned by join. The chief requirement is that a small number of widen calls
+    /// deterministically lead to Top.
+    fn widen(&self, _other: &Self, _join_condition: &AbstractDomains) -> ExpressionDomain {
+        //todo: don't get to top quite this quickly.
+        ExpressionDomain::Top
     }
 }
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -10,20 +10,100 @@ use constant_value::ConstantValue;
 /// A collection of abstract domains associated with a particular abstract value.
 /// The expression domain captures the precise expression that resulted in the abstract
 /// value. It can be used to derive any other kind of abstract domain on demand.
-#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Hash)]
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct AbstractDomains {
     pub expression_domain: ExpressionDomain,
     //todo: use cached getters to get other domains on demand
 }
 
-/// The things all abstract domain instances have to be able to do
-pub trait AbstractDomain {
-    //todo: join, is_top, is_bottom, etc.
+/// A collection of abstract domains that all represent the impossible abstract value.
+/// I.e. the corresponding set of possible concrete values is empty.
+pub const BOTTOM: AbstractDomains = AbstractDomains {
+    expression_domain: ExpressionDomain::Bottom,
+};
+
+/// A collection of abstract domains that all represent the universal abstract value.
+/// I.e. the corresponding set of possible concrete values includes every possible concrete values.
+pub const TOP: AbstractDomains = AbstractDomains {
+    expression_domain: ExpressionDomain::Top,
+};
+
+impl AbstractDomains {
+    /// True if all of the abstract domains in this collection correspond to an empty set of concrete values.
+    pub fn is_bottom(&self) -> bool {
+        self.expression_domain.is_bottom()
+    }
+
+    /// True if all of the abstract domains in this collection correspond to the set of all possible concrete values.
+    pub fn is_top(&self) -> bool {
+        self.expression_domain.is_bottom()
+    }
+
+    /// Joins all of the abstract domains in the two collections to form a single collection.
+    /// In a context where the join condition is known to be true, the result can be refined to be
+    /// just self, correspondingly if it is known to be false, the result can be refined to be just other.
+    pub fn join(
+        &self,
+        other: &AbstractDomains,
+        join_condition: &AbstractDomains,
+    ) -> AbstractDomains {
+        AbstractDomains {
+            expression_domain: self
+                .expression_domain
+                .join(&other.expression_domain, &join_condition),
+        }
+    }
+
+    /// Widen all of the abstract domains in the two collections to form a single collection.
+    pub fn widen(
+        &self,
+        other: &AbstractDomains,
+        join_condition: &AbstractDomains,
+    ) -> AbstractDomains {
+        AbstractDomains {
+            expression_domain: self
+                .expression_domain
+                .widen(&other.expression_domain, &join_condition),
+        }
+    }
+
+    /// True if for each abstract domains in the self collection, all of its concrete values are
+    /// elements of the corresponding set of the same domain in other.
+    pub fn subset(&self, other: &AbstractDomains) -> bool {
+        self.expression_domain.subset(&other.expression_domain)
+    }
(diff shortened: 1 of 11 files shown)


src/abstract_domains.rs
```diff
@@ -32,9 +112,93 @@ pub enum ExpressionDomain {
     /// that always panics.
     Bottom,
 
-    /// An expression that is a compile time constant value, such as a numeric literal or a
-    /// function.
+    /// An expression that is a compile time constant value, such as a numeric literal or a function.
     CompileTimeConstant(ConstantValue),
+
+    /// An expression that is either if_true or if_false, depending on the value of join_condition.
+    JoinedExpression {
```
zlandau: The name JoinExpression and JoinCondition is a little bit weird to me. You have an abstract domain for a condition, and based on that domain you may be able to assume if_true or if_false, but why does it matter that you get it from a join? As in, why isn't this just called Condition { condition: ... }?

It feels like maybe two things are being mixed here. In the AbstractDomain trait you have join with a condition field. I guess that makes some sense, if a lot of joins have extra information about when a join should actually be done or not. But here it sounds like it's more about if/while/etc statements. Or is it meant more generically here than branching checks?

src/abstract_value.rs
```diff
@@ -24,15 +24,67 @@ pub struct AbstractValue {
...
+
+impl AbstractValue {
+    /// True if the set of concrete values that correspond to this abstract value is empty.
+    pub fn is_bottom(&self) -> bool {
+        self.value.is_bottom()
+    }
+
+    /// True if all possible concrete values are elements of the set corresponding to this abstract value.
+    pub fn is_top(&self) -> bool {
+        self.value.is_bottom()
```
zlandau: is_top?

src/environment.rs
```diff
@@ -0,0 +1,129 @@
...
+                    debug_assert!(!val1.is_bottom());
+                    value_map = value_map.insert(
+                        p,
+                        join_or_widen(&val1, &abstract_value::BOTTOM, &join_condition),
+                    );
+                }
+            }
+        }
+        for (path, val2) in value_map2.iter() {
+            match value_map1.get(path) {
```
zlandau: if !value_map1.contains_key(path) is a little clearer to me, but whatever floats your boat.

src/visitors.rs
```diff
@@ -26,30 +28,139 @@ pub struct MirVisitor<'a, 'b: 'a, 'tcx: 'b> {
...
+        let mut iteration_count = 0;
+        while changed {
+            changed = false;
+            for bb in self.mir.basic_blocks().indices() {
+                // Merge output states of predcessors of bb
+                let mut i_state = Environment::default();
+                for pred_bb in self.mir.predecessors_for(bb).iter() {
+                    // todo: obtain join condition from out_state[pred_bb]
+                    let join_condition = &abstract_value::TOP;
+                    if iteration_count < 3 {
```
zlandau: Heh, number of iterations is a nice value to at least be defined as a const at the top or something.

src/abstract_domains.rs
```diff
@@ -10,20 +10,100 @@ use constant_value::ConstantValue;
...
 
-/// The things all abstract domain instances have to be able to do
+/// A collection of abstract domains that all represent the impossible abstract value.
+/// I.e. the corresponding set of possible concrete values is empty.
+pub const BOTTOM: AbstractDomains = AbstractDomains {
+    expression_domain: ExpressionDomain::Bottom,
+};
+
+/// A collection of abstract domains that all represent the universal abstract value.
+/// I.e. the corresponding set of possible concrete values includes every possible concrete values.
```
sblackshear: "concrete values" -> "concrete value"

src/abstract_domains.rs
```diff
@@ -10,20 +10,100 @@ use constant_value::ConstantValue;
...
+
+impl AbstractDomains {
+    /// True if all of the abstract domains in this collection correspond to an empty set of concrete values.
+    pub fn is_bottom(&self) -> bool {
+        self.expression_domain.is_bottom()
+    }
+
+    /// True if all of the abstract domains in this collection correspond to the set of all possible concrete values.
+    pub fn is_top(&self) -> bool {
+        self.expression_domain.is_bottom()
```
sblackshear: `is_top()`?

src/abstract_domains.rs
```diff
@@ -10,20 +10,100 @@ use constant_value::ConstantValue;
...
+        join_condition: &AbstractDomains,
+    ) -> AbstractDomains {
+        AbstractDomains {
+            expression_domain: self
+                .expression_domain
+                .join(&other.expression_domain, &join_condition),
+        }
+    }
+
+    /// Widen all of the abstract domains in the two collections to form a single collection.
```
sblackshear: Text about `join_condition` might also be relevant here?

src/abstract_domains.rs
```diff
@@ -10,20 +10,100 @@ use constant_value::ConstantValue;
...
+        AbstractDomains {
+            expression_domain: self
+                .expression_domain
+                .widen(&other.expression_domain, &join_condition),
+        }
+    }
+
+    /// True if for each abstract domains in the self collection, all of its concrete values are
+    /// elements of the corresponding set of the same domain in other.
+    pub fn subset(&self, other: &AbstractDomains) -> bool {
```
sblackshear: I'd prefer a more abstract name like `less_than_equal`. You could imagine a case where your domain is underapproximate and `subset` is actually the superset operation... `less_than_equal` saves you from having to commit to a particular concretization and emphasizes the lattice structure of domains.

src/abstract_domains.rs
```diff
@@ -32,9 +112,93 @@ pub enum ExpressionDomain {
...
     CompileTimeConstant(ConstantValue),
+
+    /// An expression that is either if_true or if_false, depending on the value of condition.
+    ConditionalExpression {
+        // A condition that results in a Boolean value
+        condition: Box<AbstractDomains>,
+        // The value of this expression if join_condition is true.
+        if_true: Box<ExpressionDomain>,
+        // The value of this expression if join_condition is false.
+        if_false: Box<ExpressionDomain>,
```
sblackshear: Won't `if_false` always be the negation of `if_true`? And if so, might it make sense to have `condition_expression : Box<ExpressionDomain>` and `true_branch: bool` instead to save on space?

src/abstract_domains.rs
```diff
@@ -32,9 +112,93 @@ pub enum ExpressionDomain {
...
+        }
+    }
+
+    /// True if all of the concrete values that correspond to self also correspond to other.
+    /// Note: !x.subset(y) does not imply y.subset(x).
+    fn subset(&self, other: &ExpressionDomain) -> bool {
+        match (self, other) {
+            // The empty set is a subset of every other set
+            (ExpressionDomain::Bottom, _) => true,
+            // A non empty set if not a subset of the empty set
```
sblackshear: "if" -> "is"

src/abstract_domains.rs
```diff
@@ -32,9 +112,93 @@ pub enum ExpressionDomain {
...
+                ExpressionDomain::CompileTimeConstant(cv1),
+                ExpressionDomain::CompileTimeConstant(cv2),
+            ) => cv1 == cv2,
+        }
+    }
+
+    /// Returns a domain whose corresponding set of concrete values include all of the values
+    /// corresponding to self and other.The set of values may be less precise (more inclusive) than
+    /// the set returned by join. The chief requirement is that a small number of widen calls
+    /// deterministically lead to Top.
```
sblackshear: To be pedantic: widening needs to converge to a fixed point that's a sound approximation, but not necessarily to Top. Going to Top is certainly a safe choice for implementing this more permissive specification, though :).

src/abstract_value.rs
```diff
@@ -24,15 +24,67 @@ pub struct AbstractValue {
...
+
+    /// True if all of the concrete values that correspond to self also correspond to other.
+    pub fn subset(&self, other: &AbstractValue) -> bool {
+        self.value.subset(&other.value)
+    }
+
+    /// Returns an abstract value whose corresponding set of concrete values include all of the values
+    /// corresponding to self and other. The set of values may be less precise (more inclusive) than
+    /// the set returned by join. The chief requirement is that a small number of widen calls
+    /// deterministically lead to Top.
```
sblackshear: Same nit here.

src/environment.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.join_or_widen(other, join_condition, |x, y, c| x.widen(y, c))
+    }
+
+    /// Returns an environment with a path for every entry in self and other and an associated
+    /// value that is the join or widen of self.value_at(path) and other.value_at(path).
+    fn join_or_widen(
+        &self,
+        other: &Environment,
+        join_condition: &AbstractValue,
+        join_or_widen: fn(&AbstractValue, &AbstractValue, &AbstractValue) -> AbstractValue,
```
sblackshear: The closure here feels a bit unfortunate, but I don't have a concrete suggestion for improvement.

src/environment.rs
```diff
@@ -0,0 +1,124 @@
...
+            let p = path.clone();
+            match value_map2.get(path) {
+                Some(val2) => {
+                    value_map = value_map.insert(p, join_or_widen(&val1, &val2, &join_condition));
+                }
+                None => {
+                    debug_assert!(!val1.is_bottom());
+                    value_map = value_map.insert(
+                        p,
+                        join_or_widen(&val1, &abstract_value::BOTTOM, &join_condition),
```
sblackshear: Would using `val1` yield a different result? It seems like it would be equivalent except in the case of a super-aggressive widening that goes directly to Top...

src/environment.rs
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+        for (path, val2) in value_map2.iter() {
+            if !value_map1.contains_key(path) {
+                debug_assert!(!val2.is_bottom());
+                let p = path.clone();
+                value_map = value_map.insert(
+                    p,
+                    join_or_widen(&abstract_value::BOTTOM, &val2, &join_condition),
```
sblackshear: Similar question here.

src/visitors.rs
sblackshear: At least for me, putting constants for magic numbers like this as a constant the top of the file makes it easier to see what's configurable. A comment on how 3 was chosen might also be useful.

src/abstract_domains.rs
hermanventer: I am not sure about this. I find it much easier to explain things in terms of a particular concretization and I do not currently anticipate a situation where join/widen will resulting in shrinking sets. Should things come to that, lots of things will have to change and it will be a suitable opportunity to change this name. For now, however, I find subset to be much more intuitive and explainable than less_than_or_equal.

src/abstract_domains.rs
hermanventer: No, a conditional expression is exactly what it is in C, so if_false is not the negation of if_true.
It seems the names are confusing. I'll rename if_true to consequent and if_false to alternate, if that will help.

src/abstract_domains.rs
hermanventer: Yes, and I like safe. :-)

src/abstract_value.rs
hermanventer: Same response here. :-)

src/environment.rs
hermanventer: To be pedantic, join_or_widen is not a closure but a function pointer. I considered alternatives here and did not find them as readable. On the other hand, a function pointer introduces an indirection that will be hard for the compiler to get rid of (unless it aggressively inlines method join_or_widen into methods join and widen). That said, if an extra function call is going to be a performance problem here, a lot more rewriting will be required anyway, so I'll leave this as it is for now.

src/environment.rs
hermanventer: This is a bit confusing. If we did not keep track of the join condition, val1 would do just nicely. It is not totally clear to me right now that it keeping the join condition will matter, but in general I prefer to keep as much information around as possible, for use by the expression domain.

src/environment.rs
hermanventer: Same answer as above.

src/visitors.rs
hermanventer: The comment is in line 117 below, which if not helpful if you are reading line 113, of course. I'll move the comment.

hermanventer:
The segfault was introduced by the latest nightly.

In principle it should be impossible to for a bug in Mirai to trigger a segfault since Mirai does not currently contain any unsafe code. There could, of course, be a problem with the way Mirai uses the unstable compiler API. I.e. Mirai could induce the segfault because the compiler is too trusting of the callers of its unstable APIs.

The issue, however, does not reproduce on my laptop and the CI failure produces no actionable debugging information. Pinning the build to the previous nightly makes it go away, so I'm doing that for now. I'll try the latest nightly from time to time to see if the issue resolves itself. If not, we'll have to reach out the Rust compiler team for help.

src/abstract_domains.rs
sblackshear: Got it, thanks. I don't feel strongly about the renaming.

Issue #28, opened 2019-01-10
State tracking and memory operations


hermanventer:
This is a tracking item for work related to this road map item.

hermanventer:
We now have milestones for this sort of thing.

PR #36, opened 2019-01-14
Track entry and exit conditions in block environment

hermanventer:
## Description

Track the conditions that guard control flow from one block to another and join them into path conditions that can be used to simplify (refine) abstract values that were constructed in a different context from the referring context.

Part of #28

## Type of change

## How Has This Been Tested?

Covered by existing tests. (More infrastructure is needed before the tests can become more interesting.)

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -22,13 +22,37 @@ pub const BOTTOM: AbstractDomains = AbstractDomains {
     expression_domain: ExpressionDomain::Bottom,
 };
 
+/// A collection of abstract domains that all represent the single concrete value, false.
+pub const FALSE: AbstractDomains = AbstractDomains {
+    expression_domain: ExpressionDomain::CompileTimeConstant(ConstantValue::False),
+};
+
 /// A collection of abstract domains that all represent the universal abstract value.
 /// I.e. the corresponding set of possible concrete values includes every possible concrete value.
 pub const TOP: AbstractDomains = AbstractDomains {
     expression_domain: ExpressionDomain::Top,
 };
 
+/// A collection of abstract domains that all represent the single concrete value, true.
+pub const TRUE: AbstractDomains = AbstractDomains {
+    expression_domain: ExpressionDomain::CompileTimeConstant(ConstantValue::True),
+};
+
 impl AbstractDomains {
+    /// Applies "and" to every pair of domain elements and returns the collection of results.
+    pub fn and(&self, other: &AbstractDomains) -> AbstractDomains {
+        AbstractDomains {
+            expression_domain: self.expression_domain.and(&other.expression_domain),
+        }
+    }
+
+    /// Applies "equals" to every pair of domain elements and returns the collection of results.
+    pub fn equals(&self, other: &AbstractDomains) -> AbstractDomains {
+        AbstractDomains {
+            expression_domain: self.expression_domain.equals(&other.expression_domain),
+        }
+    }
+
     /// True if all of the abstract domains in this collection correspond to an empty set of concrete values.
     pub fn is_bottom(&self) -> bool {
         self.expression_domain.is_bottom()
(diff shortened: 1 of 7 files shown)


src/abstract_domains.rs
```diff
@@ -126,9 +194,67 @@ pub enum ExpressionDomain {
...
+            }
+        } else if other.must_be_true() {
+            self.clone()
+        } else {
+            // todo: #32 more simplifications
+            ExpressionDomain::And {
+                left: box self.clone(),
+                right: box other.clone(),
+            }
+        }
```
zlandau: match (self.must_be_true(), other.must_be_true()) {
    (True, True) => ExpressionDomain::CompileTimeConstant(ConstantValue::True),
    (True, False) => other.clone(),
    (False, True) => self.clone(),
    (False, False) => ExpressionDomain::And { ... }

Makes it easier to see the logic table for must_be_trues(). But if you plan on adding a lot more than that it may or may not be helpful, up to you.

src/abstract_value.rs
```diff
@@ -102,6 +173,13 @@ impl AbstractValue {
             value: self.value.widen(&other.value, &join_condition.value),
         }
     }
+
+    /// Returns a clone of the value with the given span prepended to its provence.
+    pub fn with_provenance(&self, provenance: Span) -> AbstractValue {
+        let mut result = self.clone();
```
zlandau: I think as a general rule you want to take self and let the caller decide if they need to clone or not. But in this case the ergonomics of that might be uglier.

src/constant_value.rs
```diff
@@ -23,10 +27,25 @@ pub enum ConstantValue {
...
-    // A place holder for other kinds of constants. Eventually this goes away.
+    /// Unsigned 16 byte integer.
+    U128(u128),
+    /// The Boolean true value.
+    True,
+    /// A place holder for other kinds of constants. Eventually this goes away.
     Unimplemented,
 }
 
+impl Into<AbstractValue> for ConstantValue {
```
zlandau: According to https://doc.rust-lang.org/std/convert/trait.Into.html, you should implement From instead of Into.

src/environment.rs
```diff
@@ -20,10 +37,13 @@ impl Environment {
         let value_map = HashTrieMap::default();
         for i in 1..=num_args {
             let par_i = Path::LocalVariable { ordinal: i };
-            // todo: figure out how to get a source span for each parameter definition
             value_map.insert(par_i, abstract_value::TOP);
         }
-        Environment { value_map }
+        Environment {
```
zlandau: Environment { value_map, ..Default::default() }

Is nice in some ways, although then it doesn't force you to think about how new fields should be defined.

src/visitors.rs
```diff
@@ -114,42 +112,72 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
+                        .iter()
+                        .map(|pred_bb| {
+                            let pred_state = &out_state[pred_bb];
+                            let pred_exit_condition = pred_state.exit_conditions.get(&bb);
+                            (pred_state, pred_exit_condition)
+                        })
+                        .filter(|(_, pred_exit_condition)| pred_exit_condition.is_some())
+                        .collect();
+                    let len = predecessor_states_and_conditions.len();
+                    if len == 0 {
```
zlandau: if predecessor_states_and_conditions.is_empty()

?

src/abstract_domains.rs
hermanventer: I don't like the looks of the (False, False) option because it obscures the 3-valued logic behavior of must_be_true. I have it in mind to make must_be_true return Option<bool> to make it more painfully obvious that !must_be_true does not imply must_be_false. When I do that, I'll have another look at the match constructor for this.

src/abstract_value.rs
hermanventer: I tried to do just that, but the borrow checker was unhappy about it.

src/environment.rs
hermanventer: Since entry_conditions does not take the default value, it is slightly nicer to explicitly list all of the fields, but it is nice to know that there is such a syntax. I'll try to remember it.

PR #40, opened 2019-01-17
Add a test that exercises visit_used_copy with a structured value.

hermanventer:
## Description

Adds a test case that exercises visit_used_copy in a non trivial way, which is to say with a structured value rather than a primitive type.

## Type of change

## How Has This Been Tested?

Integration tests.


hermanventer:
It is beyond funny to see codecov reporting that ADDING a test case decreases code coverage.

PR #41, opened 2019-01-17
Implement visit_constant for integer and floating point literals.

hermanventer:
## Description

This fleshes out visit_constant to deal with numeric literals. A point to note here is that visit_constant gives out a reference to a ConstantValue, so the value has to live somewhere that outlives the visitor. That somewhere is the the constant value cache.

## Type of change

## How Has This Been Tested?

Added a test case that covers the new code.

### DIFF
--- a/src/constant_value.rs
+++ b/src/constant_value.rs
@@ -26,6 +26,10 @@ pub enum ConstantValue {
         // todo: is there some way store the def_id here if available?
         // this would not be serialized/deserialized.
     },
+    /// Signed 16 byte integer.
+    I128(i128),
+    /// 64 bit floating point, stored as a u64 to make it comparable.
+    F64(u64),
     /// The Boolean true value.
     True,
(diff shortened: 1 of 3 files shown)


src/constant_value.rs
```diff
@@ -52,15 +56,21 @@ impl ConstantValue {
 
 /// Keeps track of MIR constants that have already been mapped onto ConstantValue instances.
 pub struct ConstantValueCache {
-    cache: HashMap<DefId, ConstantValue>,
+    function_cache: HashMap<DefId, ConstantValue>,
```
zlandau: Just to be pedantic, these aren't actually caches, right? They're just lookup tables?

hermanventer: It depends on what you mean by lookup table. I do think of them as caches myself. I guess you could also call them intern tables. The basic idea is that these ConstantValues are created on demand and then cached so that future calls just look them up.

Issue #31, opened 2019-01-10
Summarize function bodies


hermanventer:
Implement the code to extract a suitable summary from the state constructed by the visitor for a function body.

PR #43, opened 2019-01-18
Bundle DefId value with function constant to help with summary caching.

hermanventer:
## Description

Include the DefId of a function in the constant value that a function reference results in. This is not serializable and function values retrieved from the summary cache will not have them. Most functions, however, should have them and this makes for a better caching story. The particular cache lookup used also registers the caller as a dependent of the function via the summary cache, so this is also providing for future incremental compilation scenarios.

Now that functions calls actually lookup the summaries of the called functions, and store the call result in the local state, it also makes sense to start making summaries have actual information in them. With this PR, return statements now update the visitor's exit_environment and the summarize function now makes use of it.

Part of fix for #28 and #31.

## Type of change

## How Has This Been Tested?

Added a new test case and inspected its debug log output to verify that the return value get's retrieved from the summary and transferred into the local state of the caller.

### DIFF
--- a/src/summaries.rs
+++ b/src/summaries.rs
@@ -38,7 +38,7 @@ use std::ops::Deref;
 ///    desirable to havoc all static variables every time such a function is called. Consequently
 ///    sound analysis is only possible one can assume that all such functions have been provided
 ///    with explicit contract functions.
-#[derive(Serialize, Deserialize, Debug, Default, Eq, PartialEq, Hash)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq, Hash)]
 pub struct Summary {
     // Conditions that should hold prior to the call.
     // Callers should substitute parameter values with argument values and simplify the results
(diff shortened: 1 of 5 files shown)


src/summaries.rs
```diff
@@ -85,18 +85,21 @@ pub struct Summary {
...
     unwind_condition: &List<AbstractValue>,
 ) -> Summary {
-    let result = None; // todo: #31 extract from exit environment
+    let result = environment.value_at(&Path::LocalVariable { ordinal: 0 });
     let side_effects: List<(Path, AbstractValue)> = List::new(); // todo: #31  extract from environment
     let unwind_side_effects: List<(Path, AbstractValue)> = List::new(); // todo: #31  extract from environment
     Summary {
         preconditions: preconditions.clone(),
-        result,
+        result: match result {
```
zlandau: Can you do

```
  result: v.map(|v| v.clone()),
```

Or something along those lines? Actually, can you clone result directly?

PR #44, opened 2019-01-18
Support & and * operators.

hermanventer:
## Description

Provide a way to model a reference to a path (location) as a value and complement that with a way to dereference a value that is a reference.

Part of #28

## Type of change

## How Has This Been Tested?

Added a new test for this. Observed from the debug log that a value flows via a reference from one local to another.


src/visitors.rs
```diff
@@ -112,6 +121,33 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
         refined_val
     }
 
+    /// For PathSelector::Deref, lookup the reference value using the qualifier, and then dereference that.
+    /// Otherwise return None.
+    fn try_to_deref(&mut self, path: &Path) -> Option<AbstractValue> {
+        match path {
```
zlandau: Not sure if this is actually an improvement or not, but at least consider:

```
if let Path::QualifiedPath { ref qualifier, ref selector } = **selector {
  if let PathSelector::Deref = selector {
    let ref_val = self.lookup_path_and_refine_result((**qualifier).clone());
     return Some(self.dereference(ref_val))
  }
}
None

hermanventer: **selector should be path and selector should be **selector, but other than that, very nice! I am still learning Box magic. Thanks.

PR #45, opened 2019-01-23
Fixed point bug fixes.

hermanventer:
## Description

Add special case code to ExpressionDomain operations to deal better with Top and Bottom.
Fix an oopsie in the "x is a subset of (condition ? consequent : alternate) x is a subset of both consequent and alternate." case.
Add == as an explicit case for ExpressionDomain::subset.
Customize various debug value formatters to make debug logs less verbose.
Remember the input state of a block after computing it from its predecessors.
Allow a qualified path to be a prefix of another qualified path.
Add some more debug log statements.

## Type of change

## How Has This Been Tested?

Existing test suffice. It would be nice if they actually failed, but were not quite there yet.
This PR started out as "Remember the input state of a block after computing it from its predecessors." because a future PR has a test that whose debug log is not as expected without it. The other changes were necessary to make the existing tests pass, given this fix, which enriches the state tracking during the fixed point computation.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -236,8 +236,13 @@ impl AbstractDomain for ExpressionDomain {
             } else {
                 other.clone()
             }
-        } else if other.as_bool_if_known().unwrap_or(false) {
+        } else if other.as_bool_if_known().unwrap_or(false)
+            || self.is_top()
+            || self.is_bottom() && other.is_bottom()
+        {
             self.clone()
+        } else if other.is_top() {
+            other.clone()
         } else {
             // todo: #32 more simplifications
             ExpressionDomain::And {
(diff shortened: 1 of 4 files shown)


src/abstract_domains.rs
```diff
@@ -258,6 +263,12 @@ impl AbstractDomain for ExpressionDomain {
 
     /// Returns an expression that is "self == other".
     fn equals(&self, other: &Self) -> Self {
+        if self == other {
+            if self.is_bottom() || self.is_top() {
+                return self.clone();
+            };
+            return ExpressionDomain::CompileTimeConstant(ConstantValue::True);
```
zlandau: Not sure I get this, why is it ConstantValue::True in this case?

hermanventer: It's a good question and I'm not totally sure of the answer. Intuitively, for any (functional) expression x, we should have x == x. Except of course if x has the runtime value of NaN. So why the special behavior for Top and Bottom? Using Bottom should always be an error, so returning Bottom seems more appropriate. Top can be a generalization of many different expressions, so self == Top and other == Top does not seem the same as x == x.

Upon further reflection this is a deeply unsatisfying answer. At the time it seemed like a good idea for equals to be in line with subset, but the operations are actually quite different. Since I don't need this change to make the fixed point terminate, I'm just going to remove it.

Thanks for noticing and speaking up about it.

hermanventer:
@sblackshear: Coming up with a small repro is probably going to take a while. Since the mirai on mirai test will stop regressions, I'm going to punt on that until I have more time.

PR #46, opened 2019-01-23
Non numeric constants

hermanventer:
## Description

Adds support for Boolean, char and str literals.

## Type of change

## How Has This Been Tested?

Added a test case.

### DIFF
--- a/src/constant_value.rs
+++ b/src/constant_value.rs
@@ -15,9 +15,11 @@ use utils::is_rust_intrinsic;
 /// value can be serialized to the persistent summary cache.
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum ConstantValue {
+    /// The primitive character type; holds a Unicode scalar value (a non-surrogate code point).
+    Char(char),
     /// The Boolean value false.
     False,
-    /// A reference to a function
+    /// A reference to a function.
     Function {
         #[serde(skip)]
         def_id: Option<DefId>,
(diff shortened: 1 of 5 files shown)


src/constant_value.rs
```diff
@@ -30,6 +32,8 @@ pub enum ConstantValue {
     I128(i128),
     /// 64 bit floating point, stored as a u64 to make it comparable.
     F64(u64),
+    /// A string literal.
+    Str(String),
```
zlandau: A Str holding a String, well that should help reduce the confusion between String and str and &str :)

I would consider naming it StringLiteral(String), otherwise rust people are going to think of &str but see String and it just gets confusing. I guess that's what happens when your analyzer for a language is written in that language, though.

src/constant_value.rs
```diff
@@ -56,21 +60,25 @@ impl ConstantValue {
 
 /// Keeps track of MIR constants that have already been mapped onto ConstantValue instances.
 pub struct ConstantValueCache {
+    char_cache: HashMap<char, ConstantValue>,
```
zlandau: type CacheMap<T> = HashMap<T, ConstantValue>;

above the struct would reduce a little bit of this repetition. But I know you prefer things to be direct and clear.

hermanventer: I'm using separate tables to avoid possible hash collision between different types of keys. Given that, HashMap<char, ConstantValue> is just a wee bit more obvious than CacheMap<char> not least because the latter name is at once non standard and quite vague/general.

src/constant_value.rs
hermanventer: I'll probably come back to this and find a way to make the parameter an actual &str. For now, the Str part strongly suggests the runtime type of the constant/literal and relates to the MIR data structures, in way that StringLiteral does not.

PR #51, opened 2019-01-25
Beef up tests.

hermanventer:
## Description

Add assertions to some more test cases. No code changes needed.

Related to #28 

## Type of change

## How Has This Been Tested?
Beefed up test cases.

### DIFF
--- a/tests/run-pass/other_constants.rs
+++ b/tests/run-pass/other_constants.rs
@@ -10,3 +10,10 @@ pub static A: bool = true;
 pub static B: bool = false;
 pub static C: char = 'A';
 pub static D: &str = "foo";
+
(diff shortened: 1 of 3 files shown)


tests/run-pass/other_constants.rs
```diff
@@ -10,3 +10,10 @@ pub static A: bool = true;
 pub static B: bool = false;
 pub static C: char = 'A';
 pub static D: &str = "foo";
+
+pub fn main() {
+    debug_assert!(A == true);
```
sblackshear: Wondering: why use `debug_assert` here rather than `assert`? Would we ever want to run the tests with the asserts compiled away?

PR #53, opened 2019-01-26
Constant folding for Rvalue::Len.

hermanventer:
## Description

Give a diagnostic if control reaches a MIR assertion statement with a false operand value. Stay silent for assert statements where the condition is the compile time constant false. This typically arises when the compiler generates blocks that are not meant to be reached, such as calling resume on a generator that has completed or panicked. While it would be nice to known when this might happen, it is beyond to current state of Mirai to do so.

One situation where an assert statement is used in a straight forward manner is during the bounds checking of an array element access. In this case, we give an error when the index is provably out of bounds.

To construct a test case of this, the visitor for the Rvalue::Len case has been enhanced to return a constant value in the case where the type of the operand is an array of a known size. I also had to add support for the < operator to abstract values.

Part of #28

## Type of change

## How Has This Been Tested?

Added test case: array_literal_out_of_bounds.rs


src/abstract_domains.rs
```diff
@@ -84,6 +84,13 @@ impl AbstractDomains {
         }
     }
 
+    /// Applies "<" to every pair of domain elements and returns the collection of results.
+    pub fn lt(&self, other: &AbstractDomains) -> AbstractDomains {
```
sblackshear: `AbstractDomains` feels like it's starting to overfit to the integer constants domain and moving in a potentially ill-typed/confusing direction as a result. `<` isn't necessarily defined for all abstract domains (same with `and`, `as_bool_if_known`, `not`, and others). 

Does it make sense to have all of these in the generic `AbstractDomains` API? Wouldn't it be safer/clearer to define separate traits for the general form of abstract domains and add additional traits for ones with extra operands like `<`?

Finally (and this is unrelated, but came to mind) the confusion of this `<` with the partial order less-than-equal is likely to arise if another abstract interpretation person reads this code :). So far, you've avoided the need to define a partial order operation less-than-equal by using a restricted version of widening that always converges to top (as opposed to always stabilizing in the more general version) + using equality/inequality to top instead. It's not clear to me that this is a good idea. In the future, as soon as someone needs to implement a more generic widening, we'll have to refactor to add partial order less-than-equal of the existing abstract domains. It might be better to just bite the bullet and do they more general thing now...

hermanventer: I am none too sure about this. In my experience the meta level should be general and it makes no sense to try and reflect concrete types into the meta types.

As for the problem of an abstract domain not supporting a particular operation: That is just saying that for that operation the domain always results in bottom.

I'd like to run with this approach for a little while longer.

src/abstract_value.rs
```diff
@@ -171,6 +171,24 @@ impl AbstractValue {
         }
     }
 
+    /// Returns an abstract value whose corresponding set of concrete values include all of the
+    /// values resulting from applying "lt" to each element of the cross product of the concrete
+    /// values or self and other.
+    pub fn lt(&self, other: &AbstractValue, expression_provenance: Option<Span>) -> AbstractValue {
+        let mut provenance = Vec::new();
+        if expression_provenance.is_some() {
```
zlandau:
```
  if let Some(expr) = expression_provenance {
    provenance.push(expr)
  }
```

PR #56, opened 2019-01-26
Array access in context needing unwind with cleanup.

hermanventer:
## Description

Adds a test case that generates an assert terminator with an unwind path.

## Type of change


tests/run-pass/array_access_with_unwind.rs
```diff
@@ -0,0 +1,14 @@
...
+//
+
+// A test that needs to do cleanup if an array access is out of bounds.
+
+pub fn foo(arr: &mut [i32], i: usize) -> String {
+    arr[i] = 123;
+    let result = String::from("foo");
+    let _e = arr[i];
+    result
+}
```
zlandau: \n

Issue #64, opened 2019-01-31
Replace debug! calls with trace! calls.


hermanventer:
## Issue

Trace output can be overwhelming so I often want to turn that off. On the other hand, since debugging support for Rust is pretty much non existent, I often need to insert some kind of print statement in the code while debugging. This is covered by the distinction between trace! and debug!, but unfortunately the code I looked at early on only ever used debug! and I just followed that.

hermanventer:
See https://crates.io/crates/log-derive.

ortem:
Hi @hermanventer! I've just heard about MIRAI. The project is very exciting and I hope it really will ever be a part of IntelliJ Rust and RLS code analysis! However, while reading the documentation I was surprised to see that CLion doesn't support debugging MIRAI ([1](https://github.com/facebookexperimental/MIRAI/blob/master/documentation/DeveloperGuide.md#debugging) and [2](https://github.com/facebookexperimental/MIRAI/blob/master/documentation/DebuggingRustc.md)).

I've just tried it out and seems like it works fine:
<img width="992" alt="Screenshot 2019-03-30 at 12 21 02" src="https://user-images.githubusercontent.com/4854600/55274221-2567e680-52e7-11e9-9bed-5b5160a7b89a.png">

All I have to do is to provide the `DYLD_LIBRARY_PATH` as specified: 
<img width="1185" alt="Screenshot 2019-03-30 at 12 35 29" src="https://user-images.githubusercontent.com/4854600/55274333-5399f600-52e8-11e9-82c1-bac7cd62c589.png">

Also,
> debugging support for Rust is pretty much non existent

I'm sorry to hear that. I'm working on new lldb formatters for Rust; in CLion, they are enabled by default and configurable via `Preferences | Build, Execution, Deployment | Debugger | Data Views | Rust`. I hope debugging support gets better, though there is no `Evaluate expression` for Rust yet. So I really want to know if something related to debugging frontend doesn't work properly or if you have some ideas to improve it (at frontend level, like formatters). If so, you can open an issue [here](https://github.com/intellij-rust/intellij-rust/issues).

PR #65, opened 2019-02-01
Compute promoted constants.

### LINKED ISSUE #34: Cache summaries for promoted constants
Promoted constants are compile time constant values that are promoted to runtime memory locations so that they may be referred to via references.

Since these constants can be complex, the MIR representation helpfully provides a function body that will compute the constant.

Before starting the analysis proper, find the promoted constants, analyze their bodies and store the Mirai abstract value from their summaries in a table that can be used by the analysis proper.

Give some though to how the code for constant A may refer to another constant B.

hermanventer:
## Description

Literal values that have their addresses taken need to be stored somewhere. For some reason or another, MIR models this by putting the code for initializing the storage in separate MIR bodies that are stored in an array that hangs of the main MIR body of a function.

For the purposes of the abstract interpreter, we'll inline the initialization code at the start of a function, by recursively invoking visit_body, taking care to reset the visitor state after each call.

Fixes #34

## Type of change

## How Has This Been Tested?

New test case.

### DIFF
--- a/src/visitors.rs
+++ b/src/visitors.rs
@@ -83,6 +83,20 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
         }
     }
 
+    /// Restores the method only state to its initial state.
+    fn reset_visitor_state(&mut self) {
+        self.check_for_errors = false;
+        self.current_environment = Environment::default();
+        self.current_location = mir::Location::START;
+        self.current_span = syntax_pos::DUMMY_SP;
+        self.exit_environment = Environment::default();
+        self.heap_addresses = HashMap::default();
+        self.inferred_preconditions = List::new();
+        self.post_conditions = List::new();
+        self.preconditions = List::new();
+        self.unwind_condition = List::new();
+    }
+
     /// Use the local and global environments to resolve Path to an abstract value.
     /// For now, promoted constants just return Top.
     fn lookup_path_and_refine_result(&mut self, path: Path) -> AbstractValue {
@@ -92,7 +106,7 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
             local_val.refine_with(&self.current_environment.entry_condition, self.current_span)
         };
         if refined_val.is_bottom() {
-            // Not found locally, so try statics and promoted constants
+            // Not found locally, so try statics.
             if let Path::StaticVariable {
                 def_id,
                 ref summary_cache_key,
@@ -109,9 +123,6 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
                     &summary
                 };
                 summary.result.clone().unwrap_or(abstract_value::TOP)
-            } else if let Path::PromotedConstant { .. } = path {
-                // todo: #34 provide a crate level environment for storing promoted constants
-                abstract_value::TOP
             } else {
                 abstract_value::TOP
             }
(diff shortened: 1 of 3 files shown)


src/visitors.rs
```diff
@@ -83,6 +83,20 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
         }
     }
 
+    /// Restores the method only state to its initial state.
+    fn reset_visitor_state(&mut self) {
```
zlandau: It kind of sucks that you have to repeat this. It would have been nice if all of these were under a substruct, and both here and in the new() you could just set it to ::default. Not sure how you feel about that.

src/visitors.rs
```diff
@@ -235,6 +248,33 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
         self.summary_cache.set_summary_for(self.def_id, summary);
     }
 
+    /// Use the visitor to compute the state corresponding to promoted constants.
+    fn promote_constants(&mut self, mut state_with_parameters: Environment) -> Environment {
+        let saved_mir = self.mir;
+        let result_root = Path::LocalVariable { ordinal: 0 };
+        let mut ordinal = 0;
+        for constant_mir in self.mir.promoted.iter() {
```
zlandau: Is .iter() necessary here? I guess for loops call IntoIter automatically but not Iter?

src/visitors.rs
```diff
@@ -235,6 +248,33 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
+            self.mir = constant_mir;
+            self.visit_body();
```
zlandau: newlines are cheap, how about adding one before and after this?

src/visitors.rs
```diff
@@ -235,6 +248,33 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
+                .exit_environment
+                .value_map
+                .iter()
+                .filter(|(p, _)| Self::path_is_rooted_by(*p, &result_root))
+            {
+                let promoted_path = Self::replace_root(&path, &result_root, promoted_root.clone());
+                state_with_parameters.update_value_at(promoted_path, value.clone());
+            }
+            self.reset_visitor_state();
+            ordinal += 1;
```
zlandau: Maybe I just can't see it, but you're not doing anything with ordinal? If you were in the loop you could just enumerate(). Actually you could do that anyway, but set the value of ordinal to the value enumerate gives you, so you could use it after the loop.

src/visitors.rs
```diff
@@ -235,6 +248,33 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
+        }
+        self.mir = saved_mir;
+        state_with_parameters
+    }
+
     /// Visits each statement in order and then visits the terminator.
     fn visit_basic_block(&mut self, bb: mir::BasicBlock) {
         debug!("visit_basic_block({:?})", bb);
```
zlandau: https://crates.io/crates/log-derive

src/visitors.rs
hermanventer: I am of mind to do that, but only when I have some spare time and an empty PR queue.

src/visitors.rs
hermanventer: So it seems.

src/visitors.rs
hermanventer: Ok.

src/visitors.rs
hermanventer: Will use enumerate.

src/visitors.rs
hermanventer: Issue #64

PR #66, opened 2019-02-01
Include side-effects into function summaries.

### LINKED ISSUE #31: Summarize function bodies

hermanventer:
## Description

This is state tracking parts of constructing and consuming function summaries.

Fixes #31 

## Type of change

## How Has This Been Tested?

New test cases.

### DIFF
--- a/src/abstract_value.rs
+++ b/src/abstract_value.rs
@@ -236,7 +236,7 @@ impl AbstractValue {
         }
     }
 
-    /// Returns a value that is simplified (refined) by using replacing parameter values
+    /// Returns a value that is simplified (refined) by replacing parameter values
     /// with their corresponding argument values. If no refinement is possible
     /// the result is simply a clone of this value.
     pub fn refine_parameters(&self, _arguments: &[AbstractValue]) -> AbstractValue {
--- a/src/abstract_value.rs
+++ b/src/abstract_value.rs
@@ -222,7 +222,7 @@ impl AbstractValue {
         }
     }
 
-    /// Returns a value that could be simplified (refined) by using the current path conditions
+    /// Returns a value that is simplified (refined) by using the current path conditions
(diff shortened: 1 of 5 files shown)


src/abstract_value.rs
```diff
@@ -236,6 +236,14 @@ impl AbstractValue {
         }
     }
 
+    /// Returns a value that is simplified (refined) by using replacing parameter values
```
zlandau: using replaced?

src/summaries.rs
```diff
@@ -85,15 +85,18 @@ pub struct Summary {
...
     preconditions: &List<AbstractValue>,
     post_conditions: &List<AbstractValue>,
     unwind_condition: &List<AbstractValue>,
+    unwind_environment: &Environment,
 ) -> Summary {
-    let result = environment.value_at(&Path::LocalVariable { ordinal: 0 });
-    let side_effects: List<(Path, AbstractValue)> = List::new(); // todo: #31  extract from environment
-    let unwind_side_effects: List<(Path, AbstractValue)> = List::new(); // todo: #31  extract from environment
+    let result = exit_environment.value_at(&Path::LocalVariable { ordinal: 0 });
+    let side_effects: List<(Path, AbstractValue)> =
```
zlandau: I guess I missed this before, but why Lists instead of Vecs?

src/abstract_value.rs
hermanventer: Fixed in a later PR.

src/summaries.rs
hermanventer: Initially I thought that these things would be part of the per block environment, so having a functional data structure would increase sharing. As things turned out, however, a Vec will do nicely instead. I'll change it.

PR #67, opened 2019-02-02
Introduce ExpressionDomain::Variable.

### LINKED ISSUE #60: Provide expression domain with a "value of this place" value to use instead of Top
We can't do Algebraic simplifications such as (x == 0) || !(x == 0) => true if we get (Top == 0) || !(Top == 0) because the value of x is unknown.

hermanventer:
## Description

When the value of a parameter is not known it is not all that useful to just set it to Top, since that prevents us from simplifying trivial expressions like x == x and more complicated expressions like (x == 0) || !(x == 0) that arise at join points and which we very much want to simplify to true.

To solve this, the expression domain now has a value for "the value of local n". Since this is not Top, we can conclude that two such values with the same n are in fact equal. Of course, this may not be true if the local contains a floating point NaN value, so the value additionally tracks the type of the local to allow us to special case this.

Fixes #60 

## Type of change

## How Has This Been Tested?

New test case.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -244,6 +244,41 @@ pub enum ExpressionDomain {
 
     /// The corresponding concrete value is the runtime address of location identified by the path.
     Reference(Path),
+
+    /// The unknown value of a place in memory.
+    /// This is distinct from Top in that we known something: the place and the type.
+    /// This is a useful distinction because it allows us to simplify some expressions
+    /// like x == x. The type is needed to prevent this particular optimization if
+    /// the variable is a floating point number that could be NaN.
+    Variable {
+        path: Box<Path>,
+        var_type: ExpressionType,
+    },
+}
+
+/// The type of a place in memory, as understood by MIR.
+/// For now, we are only really interested to distinguish between
+/// floating point values and other values, because NaN != NaN.
+/// In the future the other distinctions may be helpful to SMT solvers.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum ExpressionType {
+    Bool,
+    Char,
+    F32,
+    F64,
+    I8,
+    I16,
+    I32,
+    I64,
+    I128,
+    Isize,
+    NonPrimitive,
+    U8,
+    U16,
+    U32,
+    U64,
+    U128,
+    Usize,
 }
 
 impl AbstractDomain for ExpressionDomain {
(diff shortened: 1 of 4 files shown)


src/abstract_domains.rs
```diff
@@ -244,6 +244,41 @@ pub enum ExpressionDomain {
...
+/// The type of a place in memory, as understood by MIR.
+/// For now, we are only really interested to distinguish between
+/// floating point values and other values, because NaN != NaN.
+/// In the future the other distinctions may be helpful to SMT solvers.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum ExpressionType {
+    Bool,
+    Char,
+    F32,
+    F64,
```
zlandau: Careful if F128 ends up getting in.

src/abstract_domains.rs
```diff
@@ -291,20 +326,46 @@ impl AbstractDomain for ExpressionDomain {
...
+                    var_type: t1,
+                },
+                ExpressionDomain::Variable {
+                    path: p2,
+                    var_type: t2,
+                },
+            ) => {
+                if (**p1) == **p2 {
+                    match (t1, t2) {
+                        (ExpressionType::F32, _)
```
zlandau: These functions are on the cusp of getting too long and intended for me. Consider having some functions like is_floating_point_expression or whatever.

src/visitors.rs
```diff
@@ -144,8 +176,7 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
         }
         // The entry block has no predecessors and its initial state is the function parameters
         // as well any promoted constants.
-        let state_with_parameters = Environment::with_parameters(self.mir.arg_count);
```
zlandau: What happened to self.mir.arg_count? Do you still need that?

src/abstract_domains.rs
hermanventer: The type checker will complain when that happens.

src/abstract_domains.rs
hermanventer: Ooo, they are going to get a lot longer before long. Let me think it over a little later on.

src/visitors.rs
hermanventer: I still need self.mir. The arg_count I don't need, but mir is not my data structure.

PR #71, opened 2019-02-08
Implement refine_parameters

hermanventer:
## Description

Specialize function summaries to particular call sites by walking the expressions in the (copy of the) summary and replacing all occurrences of parameter references with the actual argument values at the call site.

## Type of change

## How Has This Been Tested?

Modified an existing test case with an expression that has a parameter reference. Getting full code coverage would mean writing every possible kind of expression, so I'm putting that off for now.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -854,11 +854,11 @@ impl AbstractDomain for ExpressionDomain {
     fn and(&self, other: &Self) -> Self {
         let self_bool = self.as_bool_if_known();
         if let Some(false) = self_bool {
-            return ExpressionDomain::CompileTimeConstant(ConstantValue::False);
+            return false.into();
         };
         let other_bool = other.as_bool_if_known();
         if let Some(false) = other_bool {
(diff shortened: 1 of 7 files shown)


src/abstract_domains.rs
```diff
@@ -1644,6 +1654,125 @@ impl AbstractDomain for ExpressionDomain {
...
+                .sub(&right.refine_parameters(arguments)),
+            ExpressionDomain::SubOverflows {
+                left,
+                right,
+                result_type,
+            } => left
+                .refine_parameters(arguments)
+                .sub_overflows(&right.refine_parameters(arguments), result_type.clone()),
+            ExpressionDomain::Variable { path, .. } => match **path {
+                Path::LocalVariable { ordinal } if 0 < ordinal && ordinal <= arguments.len() => {
```
zlandau: Did you mean use Yoda ordering on the '0 < ordinal'?

hermanventer: Yes, I like to use 0 < x & x < n as an approximation for 0 < x < n.

PR #72, opened 2019-02-08
Add some basic simplifications

hermanventer:
## Description

Expand the expression factories to do some more basic simplifications. These simplifications are effective for the expressions that arise in the entry conditions of basic blocks because of the control flow that Rustc generates for short circuit expressions, as well as the expressions that result at the join points of if statements and switch statements.

## Type of change

## How Has This Been Tested?

Uncommented some short circuit expressions in assert statements.

### DIFF
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -854,11 +854,11 @@ impl AbstractDomain for ExpressionDomain {
     fn and(&self, other: &Self) -> Self {
         let self_bool = self.as_bool_if_known();
         if let Some(false) = self_bool {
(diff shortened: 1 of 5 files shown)


src/abstract_domains.rs
```diff
@@ -854,11 +854,11 @@ impl AbstractDomain for ExpressionDomain {
     fn and(&self, other: &Self) -> Self {
         let self_bool = self.as_bool_if_known();
         if let Some(false) = self_bool {
-            return ExpressionDomain::CompileTimeConstant(ConstantValue::False);
+            return false.into();
```
zlandau: Ahhhh these are nice.

PR #76, opened 2019-02-12
Infer preconditions

hermanventer:
## Description

A false assertion or panic that may be reachable in a function is not necessarily a bug if the function requires its callers to supply parameter values that will ensure that such conditions never arise.

Consequently, instead of reporting such issues as potential errors in the function, the path conditions that could lead to the error conditions are inverted and added to as inferred preconditions of the function.

In a future PR the call sites will be enhanced to check the preconditions of the called functions.

Note that this scheme of doing things breaks down for functions called indirectly via traits and for public functions that are called by code that is not being analyzed. Such functions will need programmer supplied preconditions that are enforced at runtime. Al that will have to wait a bit longer.

## Type of change

## How Has This Been Tested?

Modified some existing tests. I also run Mirai on itself on my development machine. We no longer do that on Travis since it uses too much memory.

### DIFF
--- a/src/visitors.rs
+++ b/src/visitors.rs
@@ -560,11 +560,25 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
                 .as_bool_if_known()
                 .unwrap_or(false)
         {
-            let span = self.current_span;
-            let mut err = self
-                .session
-                .struct_span_warn(span, "Execution might panic.");
-            (self.emit_diagnostic)(&mut err, &mut self.buffered_diagnostics);
+            //todo: use theorem prover to make sure that the entry condition is not known to be false.
+            if self
+                .current_environment
+                .entry_condition
+                .as_bool_if_known()
+                .unwrap_or(false)
+            {
+                let span = self.current_span;
+                let mut err = self
+                    .session
+                    .struct_span_warn(span, "Execution might panic.");
+                (self.emit_diagnostic)(&mut err, &mut self.buffered_diagnostics);
+            } else {
+                self.preconditions.push(
+                    self.current_environment
+                        .entry_condition
+                        .not(Some(self.current_span)),
+                );
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/visitors.rs
```diff
@@ -679,21 +693,65 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
-                    span,
-                    "Control might reach a call to std::intrinsics::unreachable.",
-                );
-                (self.emit_diagnostic)(&mut err, &mut self.buffered_diagnostics);
+                //todo: use theorem prover to make sure that the entry condition is not known to be false.
+                if self
+                    .current_environment
+                    .entry_condition
+                    .as_bool_if_known()
+                    .unwrap_or(false)
```
zlandau: You have this .as_bool_if_known().unwrap_or(false) repeated so many times, it might be nice if there was a helper for it. Maybe you can have an OptionExt trait that adds:

  .as_bool_or_false()

It just feels a little nicer to me. I don't know, maybe it's not worth it. You can ponder it.

src/visitors.rs
```diff
@@ -679,21 +693,65 @@ impl<'a, 'b: 'a, 'tcx: 'b> MirVisitor<'a, 'b, 'tcx> {
...
+                };
+                // Since the assertion was conjoined into the entry condition because of the logic
+                // of debug_assert!, and this condition was simplified, we cannot distinguish
+                // between the case of maybe reaching a definitely false assertion from the case of
+                // definitely reaching a maybe false assertion.
+                // We therefore report both cases, even though the first would be a candidate for being
+                // an inferred precondition, rather than a diagnostic.
+                //todo: we need our own library for contract assertions so that we can distinguish
+                // such cases.
+                if msg.starts_with("assertion failed:")
```
zlandau: Wah? Checking string values and replacing them seems weird. I'm too lazy to check to see where this string is set, but don't you control this? Can't you pass something other than a string?

hermanventer: I'm afraid not. The string is generated by Rustc. I am not happy about this either, but this is the best I can do at present.

src/visitors.rs
hermanventer: It feels a little nicer to me too, but upon reflection the name is not satisfying. It also bears noting that the rather painful sequence it replaces has a desirable function in as much as it slows down the reader and forces them to remember that that logic is ternary, not binary.

PR #78, opened 2019-02-13
Check preconditions

hermanventer:
## Description

The crux of this PR is to actually check the inferred preconditions of a function at the point of call. Doing that requires a new kind of refinement operation. It also made visit_call so large that it is now broken up into many helper functions. Reviewing that code caused a new todo and some tweaks.

We now also have the first instance of a diagnostic with related locations. This uncovered a bug in the test runner. Debugging that made the random failures of the test runner more painful, so I looked into that also. That turns out to be a problem of running multiple invocations of LLVM in the same process. The test runner now arranges for the compiler controller to stop the compilation after Mirai has run. So now cargo test works every time again (on the Mac). Yay!!!

## Type of change

## How Has This Been Tested?

Added a new test case. Ran validate.sh, locally.

### DIFF
--- a/src/summaries.rs
+++ b/src/summaries.rs
@@ -44,7 +44,9 @@ pub struct Summary {
     // under the current path condition. Any values that do not simplify to true will require the
     // caller to either generate an error message or to add a precondition to its own summary that
     // will be sufficient to ensure that all of the preconditions in this summary are met.
-    pub preconditions: Vec<AbstractValue>,
+    // The string value bundled with the condition is the message that details what would go
+    // wrong at runtime if the precondition is not satisfied by the caller.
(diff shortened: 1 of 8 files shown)


src/summaries.rs
```diff
@@ -44,7 +44,7 @@ pub struct Summary {
     // under the current path condition. Any values that do not simplify to true will require the
     // caller to either generate an error message or to add a precondition to its own summary that
     // will be sufficient to ensure that all of the preconditions in this summary are met.
-    pub preconditions: Vec<AbstractValue>,
+    pub preconditions: Vec<(AbstractValue, String)>,
```
zlandau: Can you add a comment about what String holds here?

PR #81, opened 2019-02-19
Add an interval domain

### LINKED ISSUE #62: Provide an interval domain
## Issue

Create an abstract domain for tracking intervals such as [0, ..., n] and so on.

hermanventer:
## Description

Adds an interval domain. The domain is constructed upon demand when checking for arithmetic overflow and index in range conditions. It is not complete, but only implements operations that might reasonably be expected to arise in real code.

Bounded intervals have to be proper subsets of the open interval (i128::MIN, ..., i128:MAX). The min and max values are used to denote that the lower/upper bound, respectively, of an interval is unbounded (infinite).  

Fixes #62

## Type of change

## How Has This Been Tested?

Added new test cases. Mirai on Mirai.

### DIFF
--- a/src/interval_domain.rs
+++ b/src/interval_domain.rs
@@ -1,0 +1,253 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+use expression::ExpressionType::{self, *};
+use std::cmp;
+use std::convert::TryFrom;
+
+/// An element of the Interval domain is a range of i128 numbers denoted by a lower bound and
+/// upper bound. A lower bound of std::i128::MIN denotes -infinity and an upper bound of
+/// std::i128::MAX denotes +infinity.
+/// Interval domain elements are constructed on demand from AbstractDomain expressions.
+/// They are most useful for checking if an array index is within bounds.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialOrd, PartialEq, Hash)]
+pub struct IntervalDomain {
+    lower_bound: i128,
+    upper_bound: i128,
+}
+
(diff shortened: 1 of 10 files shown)


src/interval_domain.rs
```diff
@@ -0,0 +1,254 @@
...
+use std::cmp;
+use std::convert::TryFrom;
+
+/// An element of the Interval domain is a range of i128 numbers denoted by a lower bound and
+/// upper bound. A lower bound of std::i128::MIN denotes -infinity and an upper bound of
+/// std::i128::MAX denotes +infinity.
+/// Interval domain elements are constructed on demand from AbstractDomain expressions.
+/// They are most useful for checking if an array index is within bounds.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialOrd, PartialEq, Hash)]
+pub struct IntervalDomain {
```
zlandau: It seems like
 
  type Value = Min | Number i128 | Max;
  type Type =  Top | { lower_bound: Value, upper_bound: Value } | Top

is more 'correct', but using things like MIN and MAX for Value and lower_bound: 1, upper_bound: 0 for bottom is just to make operations more convenient?

src/interval_domain.rs
```diff
@@ -0,0 +1,254 @@
...
+            lower_bound: i,
+            upper_bound: i,
+        }
+    }
+}
+
+impl From<u128> for IntervalDomain {
+    fn from(u: u128) -> IntervalDomain {
+        let i = i128::try_from(u);
+        if i.is_ok() {
```
zlandau: Personally I'd probably do:

  if let i128::try_from(i) = u {
    ...
  } else {
    ...
  }

src/interval_domain.rs
hermanventer: Once you introduce more cases, everything pretty much explodes.

PR #85, opened 2019-02-27
[WIP] Hook up Z3 SMT Solver

hermanventer:
## Description

This is not for merging, but to illustrate how an SMT solver can get integrated into MIRAI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libssl-dev
 language: rust
 rust:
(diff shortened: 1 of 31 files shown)


hermanventer:
Sadly the crate providing the Rust bindings does not include Z3. Also, I have yet to figure out how to make things work on Linux. Ideally, I would just include or reference pre-built binaries, since building Z3 takes a long time.

PR #87, opened 2019-03-04
Outer fixed point

hermanventer:
## Description

If function a depends on function b, which succeeds its in source order, then a is being analyzed without the benefit of a summary for b, which could lead to both false positives (because the lack of knowledge about the result (post conditions) of b may lead to overly pessimistic assumptions) and false negatives (because a's call to be may violate unknown preconditions of b).

In order to avoid this, any function that depends on another function whose summary is unknown (or incomplete) has to get re-analyzed. This process has to be repeated until all summaries are complete. We regard a summary as complete when it does not change after being recomputed with updated summaries for the functions it calls. 

This represents a fixed point, usually referred to as the "outer fixed point".

In order to safely compare a new summary with a previous summary, the collections in summaries should be sorted. That requires AbtractValues to be comparable, which requires abstract operations such as gt and ge to be renamed so that they do not clash with the corresponding concrete operations on AbstractValues in the interpreter.

Finally, note that appropriate widening is required for the fixed point loop to converge. This is not supplied by this PR, so instyead there is a new k-limit that terminates the fixed point loop before it causes too much additional runtime. Even so, the runtime of Mirai now more than doubles when it is run on itself.

## Type of change

## How Has This Been Tested?
Modified an existing test so that a function calls another that has not yet been analyzed in round 1.
Ran MIRAI on MIRAI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libssl-dev
 language: rust
 rust:
-- nightly-2019-02-27
+- nightly
 cache: cargo
 
 before_script:
--- a/src/abstract_domains.rs
+++ b/src/abstract_domains.rs
@@ -18,7 +18,7 @@ use syntax::ast;
 
 /// Basically, this domain is a structured container for other domains. It is also the only
 /// client for the other domains.
-#[derive(Serialize, Deserialize, Clone, Eq)]
+#[derive(Serialize, Deserialize, Clone, Eq, Ord, PartialOrd)]
 pub struct AbstractDomain {
     // todo: make this private
     // This is not a domain element, but a representation of how this instance has been constructed.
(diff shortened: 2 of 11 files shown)


int3:
> Even so, the runtime of Mirai now more than doubles when it is run on itself.

Have you looked at using a weak-topological-ordering based fixpoint iterator to optimize performance? @arnaudvenet implemented it in SPARTA [here](https://github.com/facebookincubator/SPARTA/blob/master/include/WeakTopologicalOrdering.h) and [here](https://github.com/facebookincubator/SPARTA/blob/master/include/MonotonicFixpointIterator.h). It's based on [this paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.89.8183&rep=rep1&type=pdf). I played around with [porting the WTO algorithm to Rust](https://github.com/int3/MIRAI/commit/c0a3ce0a99c59009825b11fd590c8fb199bb5e00), though I haven't gotten around to integrating it with MIRAI's fixpoint iterator. I could look at doing that sometime if there's interest :)

hermanventer:
I haven't had the opportunity to look at SPARTA yet, but I hear good things about it. It would be super interesting to make some use of it in MIRAI. If you can help out with that it would be magic. Thanks!

PR #91, opened 2019-03-07
Support array constants

hermanventer:
## Description

Somewhat surprisingly, MIR has two ways to encode structured compile time constants: 1) As MIR instructions in a promoted constants body and 2) As a string of bytes that contain a runtime image of the constant (along with a map of undefined bytes and a map of pointers to relocate).

To add insult to injury, a single source literal can be encoded both ways. Oh well.

This PR deals with array literals that are compile time constants and that are encoded as byte strings. Things are a bit more involved than it may seem because you can refer to an array directly or you can refer to a slice of the array. The former only seems to happen if the array is in fact a "byte string" such as b"a", while the latter happens when a fixed length array is referenced via an unsized reference (i.e. when a runtime length is added to the reference by referring to a slice of the array).

## Type of change

## How Has This Been Tested?
cargo test and validate.sh
New test cases.

### DIFF
--- a/src/visitors.rs
+++ b/src/visitors.rs
@@ -715,6 +715,8 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                 .refine_parameters(actual_args)
                 .refine_paths(&mut self.current_environment)
                 .refine_with(&self.current_environment.entry_condition, self.current_span);
+            //todo: if refined_precondition is a Variable, look it up
+            // or perhaps pass in &mut self.current_environment
             let (refined_precondition_as_bool, entry_cond_as_bool) =
                 self.check_condition_value_and_reachability(&refined_precondition);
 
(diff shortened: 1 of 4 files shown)


src/visitors.rs
```diff
@@ -1882,26 +2164,36 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
...
-                        TyKind::Adt(..) => base_ty,
-                        TyKind::RawPtr(ty_and_mut) => &ty_and_mut.ty.sty,
-                        TyKind::Ref(_, ty, _) => &ty.sty,
-                        _ => unreachable!(),
+                mir::ProjectionElem::Deref => match base_ty {
+                    TyKind::Adt(..) => base_ty,
+                    TyKind::RawPtr(ty_and_mut) => &ty_and_mut.ty.sty,
+                    TyKind::Ref(_, ty, _) => &ty.sty,
+                    _ => {
+                        println!("span: {:?}", self.current_span);
```
zlandau: Did you mean to leave these printlns in here?

hermanventer: Yes, I am by no means convinced that these cases are actually unreachable, so if they actually occur, I'd like to see a bit more information than just there is a bug in Mirai. I could put them all in a string message and pass it to unreachable, but that seems a little less readable to me.

bzvl: Why not use the logger then?

PR #93, opened 2019-03-11
Limit the number of inferred preconditions that a summary can have

hermanventer:
## Description

Large self recursive functions, such as those found in Mirai can end up with many hundreds of inferred preconditions and this is one source of noticeable slowdown for Mirai.

This change limits the number of preconditions that can be pushed for any function. Once the limit is reached, warnings are issued at the source of a panic, rather than at call sites. This increases false positives because some infeasible paths will now be treated as feasible.

Imposing this restriction claws back about 1 minute of execution time when running Mirai on Mirai. 

## Type of change

## How Has This Been Tested?
cargo test and validate.sh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libssl-dev
 language: rust
 rust:
-- nightly
+- nightly-2019-03-09
 cache: cargo
 
 before_script:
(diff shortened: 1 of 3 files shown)


hermanventer:
For any given source version, the results of Mirai should be stable. Also, there is no plan for base lines. I'll keep logging in mind. For now it is kind of obvious when this happens:  a private method will have diagnostics in situations where you'd expect a precondition to get pushed. It will only happen in large recursive functions like those in MIRAI.

PR #95, opened 2019-03-13
Use the new Rust compiler interface

hermanventer:
## Description

Updates MIRAI to use the new compiler plug-in interface for analyzers.

The new interface required a rewrite of the diagnostic handling code. In the rewrite the problem with duplicate messages because of the outer fixed point loop is now resolved. As a side effect, there is no output from the compiler until it has finished the run.

## Type of change

## How Has This Been Tested?
cargo test and validate.sh


hermanventer:
It has been worked on since December but only made it into nightly very recently.

PR #101, opened 2019-03-22
Report fewer related locations for promoted preconditions.

hermanventer:
## Description

When a panic point in a function is promoted into a precondition (because the function containing it is not public) and made its caller's problem, the number of source locations that are reported quickly become overwhelming and confusing because it includes not just the "call stack" but also every location that leads to a condition that is on a path to the panic point.

This PR reduces the related locations to be just the call stack, so it will be the call to the function with the precondition that may not hold and then the location where the panic may occur in the called function (which itself may be a call to another function and so on).

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -523,6 +523,13 @@ impl AbstractValue {
         }
     }
 
+    /// Returns a clone of the value with the given span replacing its provenance.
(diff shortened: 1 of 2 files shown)


zlandau:
I assume the new assertions weren't meant to be part of this review?

zlandau:
Ah disregard comment, I think I was just viewing it wrong.

PR #103, opened 2019-03-26
Fix initial state of out of order non root blocks

hermanventer:
## Description

Instead of marking the precondition of a block with no visited predecessors as False (unreachable) mark it as Top (unknown). Furthermore, if the visited predecessors all concluded that the block will not be reached, then don't visit the block. 

## Type of change

## How Has This Been Tested?
cargo test; validate.sh


zlandau:
I guess I don't understand the 'why' here. Normally I would expect 'nothing is yet known' to be bottom, not top. So why the switch?

hermanventer:
Top means "nothing is known", Bottom means "known to be impossible". You may ask why not just mark the precondition as True? Well, that will probably work, but it implies certainty, so it is as wrong as False. Right now, Top seems correct to me.

PR #104, opened 2019-03-26
Limit the number of diagnostics given for self recursive calls

hermanventer:
## Description

Self recursive functions that do not satisfy their own preconditions should not add additional preconditions to make the problem of their callers, since, well, they are their callers already.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh


checker/src/visitors.rs
```diff
@@ -44,6 +44,7 @@ pub struct MirVisitor<'a, 'b: 'a, 'tcx: 'b, E> {
     summary_cache: &'a mut PersistentSummaryCache<'b, 'tcx>,
     smt_solver: &'a mut dyn SmtSolver<E>,
 
+    already_report_errors_for_call_to: HashSet<AbstractValue>,
```
zlandau: super_nit: shouldn't this be 'already_reportED_errors_for_call_to' ?

PR #116, opened 2019-04-03
Introduce a form of time-out to prevent MIRAI from taking too long.

hermanventer:
## Description

A few functions consume most of the analysis time and makes the overall analysis time for MIRAI on MIRAI unacceptably long. For the time being, just ignore (and log) functions that take longer than 3 seconds to analyze.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -42,6 +42,11 @@ use std::ops::Deref;
 ///    with explicit contract functions.
 #[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq, Hash)]
 pub struct Summary {
+    /// The number of seconds that the analyzer used to construct this summary.
+    /// If it is >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY then this summary is not definitive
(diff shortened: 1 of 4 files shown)


checker/src/summaries.rs
```diff
@@ -42,6 +42,11 @@ use std::ops::Deref;
 ///    with explicit contract functions.
 #[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq, Hash)]
 pub struct Summary {
+    /// The number of seconds that the analyzer used to construct this summary.
+    /// If it is >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY then this summary is not definitive
+    /// because the analyzer will omit it from the outer fixed point loop.
+    pub analysis_time_in_seconds: u64,
```
zlandau: Any particular reason you want to use u64 instead of something like Duration?

hermanventer: Duration seems a bit like overkill and it does not support a default formatter, but it is a fair point. I keep it in mind for the future.

PR #118, opened 2019-04-03
Fix path propagation and refinement bugs.

hermanventer:
## Description

A number of small fixes to path propagation and refinement. The new test case exercises most of these, but mostly this comes from debugging MIRAI on MIRAI.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -847,6 +847,18 @@ impl Path {
                         // can refine this value.
                     }
                 }
+            } else {
+                // The qualifier does not match a value in the environment, but parts of it might.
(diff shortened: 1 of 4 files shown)


checker/src/abstract_value.rs
```diff
@@ -847,6 +847,15 @@ impl Path {
                         // can refine this value.
                     }
                 }
+            } else {
```
zlandau: nit: I have to scroll back 20 pages to see what this is an else for, a comment about it would have saved me 2.5s of my life.

hermanventer: Comment duly added.

PR #120, opened 2019-04-04
Turn call to core.slice.len into lookup of ArrayLength path.

hermanventer:
## Description

The core.slice.len function is particularly problematic for MIRAI in its current state. Firstly it is a standard method, and building summaries for the standard library is still a ways away (see #2). Secondly, it is a trait method so there is no one body that provides a summary for it. Thirdly, it is generic, which is something that we generally don't need to support. Lastly, it is actually impossible to write a summary for it in Rust itself, since there is not syntactic operator that corresponds to getting the length of a slice.

Unfortunately there is one place in the MIRAI code base where using this function in unavoidable. So, in order to get MIRAI to check MIRAI, this pull requests add special handling for calls to this function. In effect, the summary for this call is now built into MIRAI.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Note that this PR by itself is not enough for the test case to have no diagnostics.

### DIFF
--- a/checker/src/constant_domain.rs
+++ b/checker/src/constant_domain.rs
@@ -637,6 +637,7 @@ pub struct ConstantValueCache {
     i128_cache: HashMap<i128, ConstantDomain>,
     u128_cache: HashMap<u128, ConstantDomain>,
     str_cache: HashMap<String, ConstantDomain>,
+    core_slice_len_function: Option<ConstantDomain>,
     mirai_assume_function: Option<ConstantDomain>,
     mirai_verify_function: Option<ConstantDomain>,
     std_panicking_panic_function: Option<ConstantDomain>,
@@ -653,6 +654,7 @@ impl ConstantValueCache {
             i128_cache: HashMap::default(),
             u128_cache: HashMap::default(),
             str_cache: HashMap::default(),
+            core_slice_len_function: None,
             mirai_assume_function: None,
             mirai_verify_function: None,
             std_panicking_panic_function: None,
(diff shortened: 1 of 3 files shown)


checker/src/visitors.rs
```diff
@@ -726,6 +726,19 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                 }
                 return;
             }
+            let result: AbstractValue = self.try_to_inline_standard_ops_func(fun, &actual_args);
```
zlandau: This code block feels super specific to get such a prominent place in visit_call(). Can you reasonable abstract it out? Mostly I'm just concerned that things like this made it hard to read visit_call() and understand generally what it is doing.

checker/src/constant_domain.rs
```diff
@@ -637,6 +637,7 @@ pub struct ConstantValueCache {
     i128_cache: HashMap<i128, ConstantDomain>,
     u128_cache: HashMap<u128, ConstantDomain>,
     str_cache: HashMap<String, ConstantDomain>,
+    core_slice_len_function: Option<ConstantDomain>,
```
zlandau: Ok for now, but if you end up having more of these you should use some container type.

hermanventer: #125

checker/src/visitors.rs
hermanventer: I have not found a better way of doing it.

PR #123, opened 2019-04-05
Fix some problems reported by Mirai on Mirai

hermanventer:
## Description

Running MIRAI on itself uncovered some undocumented assumptions. They are now explicit. 

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -7,6 +7,7 @@ use crate::abstract_domains::{self, AbstractDomain};
 use crate::constant_domain::ConstantDomain;
 use crate::environment::Environment;
 use crate::expression::{Expression, ExpressionType};
+use crate::k_limits;
 
 use rustc::hir::def_id::DefId;
 use std::collections::HashSet;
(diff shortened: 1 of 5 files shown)


checker/src/abstract_value.rs
```diff
@@ -762,7 +763,7 @@ impl Path {
     pub fn refine_parameters(&self, arguments: &[(Path, AbstractValue)]) -> Path {
         match self {
             Path::LocalVariable { ordinal } if 0 < *ordinal && *ordinal <= arguments.len() => {
-                arguments[ordinal - 1].0.clone()
+                arguments[*ordinal - 1].0.clone()
```
hermanventer: Not so much an undocumented assumption as a rather subtle call to a virtual function when a simple subtraction is expected.

checker/src/constant_domain.rs
```diff
@@ -274,7 +274,11 @@ impl ConstantDomain {
                 if *val2 == 0 {
                     ConstantDomain::Bottom
                 } else {
-                    ConstantDomain::I128(*val1 / *val2)
+                    let result = val1.checked_div(*val2);
```
zlandau: ` val1.checked_div(*val2).map(ConstantDomain::I128).map_err(ConstantDomain::Bottom)`

does that work?

hermanventer: Nope, but I guess you really meant:
`val1.checked_div(*val2).map(ConstantDomain::I128).unwrap_or(ConstantDomain::Bottom)`

hermanventer:
Not easy, no. You'll have to build the latest greatest MIRAI including all the pull requests,, install that in cargo (as per the validate.sh), install Z3, then switch back to before this commit and then run Mirai over itself (as per validate.sh).

hermanventer:
Oh, and you'll have to not use Linux.

Issue #125, opened 2019-04-05
Precompute if a function is well known


hermanventer:
## Issue

Some functions from the standard libraries (and from mirai-annotations) are well known to MIRAI and calls to them have semantics that cannot be expressed in summaries.

Currently there is an elaborate and not very scalable scheme for checking if a function is well known. Basically, it boils down to checking if the summary cache key is well known. Since this is an expensive check it needs to be cached.

Currently that is done in ConstantValueCache, but many expensive checks may happen before caching succeeds. While it is conceivable that the total number of such checks will be smaller than the total number of function values created during the lifetime of ConstantValueCache, the current scheme does not guarantee a bound to the number of checks.

An alternative scheme is to set aside a field in each function object that is an enum that lists all of the well known functions (and has an "other" value for non well known functions). This field must then be initialized (via the expensive summary cache key check) when a function is constructed.

Code sites that want to check if a function is a particular well known function, will then just have to check the value of this enum.

PR #126, opened 2019-04-05
Add precondition macros to MIRAI Annotations

hermanventer:
## Description

Preconditions are assumptions inside the functions that contain them, but need to be verified by callers of the functions that contain them. This makes them distinct from assume and verify and requires a separate set of macros to be added to mirai-annotations.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
New test case, modified old test case.

### DIFF
--- a/checker/src/constant_domain.rs
+++ b/checker/src/constant_domain.rs
@@ -641,6 +641,7 @@ pub struct ConstantValueCache {
     str_cache: HashMap<String, ConstantDomain>,
     core_slice_len_function: Option<ConstantDomain>,
     mirai_assume_function: Option<ConstantDomain>,
+    mirai_precondition_function: Option<ConstantDomain>,
     mirai_verify_function: Option<ConstantDomain>,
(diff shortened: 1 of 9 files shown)


checker/src/constant_domain.rs
```diff
@@ -770,6 +772,27 @@ impl ConstantValueCache {
         })
     }
 
+    /// Does an expensive check to see if the given function is mirai_annotations.mirai_precondition.
+    /// Once it finds the function it caches it so that subsequent checks are cheaper.
+    pub fn check_if_mirai_precondition_function(&mut self, fun: &ConstantDomain) -> bool {
```
zlandau: Ah, this cached summary key pattern shows up multiple times :) Can you abstract the code below? It's a bit complex, and if it's going to be repeated then it'd be nice to have some code sharing.

hermanventer: #125

PR #127, opened 2019-04-05
Treat the assert! macro as a way of stating a precondition.

hermanventer:
## Description

The panic! macro, sometimes in the guise of assert!, is sometimes (perhaps even often) used to document preconditions in existing code (in the wild, a.k.a. crates.io).

Given a diagnostic for every panic! call results in too many false positives, so they are now silently promoted to preconditions. These preconditions are in turn silently promoted by callers until the caller is public, in which case it is loudly promoted.

This still results in false positives, but this seems the best trade-off between false positives and false negatives that can be achieved without programmer annotation or very sophisticated whole program analysis.

Also in this PR, a small change that allows code generation to happen after MIRAI is done with its analysis. Although it is a non-goal for MIRAI to be the normal compiler to use in every day builds, it turns out that build scripts (build.rs) need to be compiled and run as part of a MIRAI build.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/callbacks.rs
+++ b/checker/src/callbacks.rs
@@ -183,6 +183,6 @@ impl rustc_driver::Callbacks for MiraiCallbacks {
             }
             info!("done with analysis");
         });
-        false // Do not proceed to code generation. MIRAI is only a checker.
+        true // Although MIRAI is only a checker we still need code generation for build scripts.
     }
 }
--- a/checker/src/visitors.rs
+++ b/checker/src/visitors.rs
@@ -882,18 +882,27 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                     && !refined_precondition_as_bool.unwrap_or(true)
                 {
                     // We always get here if the function is called, and the precondition is always
-                    // false, so complain loudly.
-                    if !self
-                        .already_report_errors_for_call_to
-                        .contains(&func_to_call)
-                    {
-                        self.emit_diagnostic_for_precondition(precondition, &message);
-                        self.already_report_errors_for_call_to
-                            .insert(func_to_call.clone());
+                    // false, so it may seem to make sense to just complain loudly here and not
+                    // promote the precondition, which can never be satisfied by the caller.
+                    // It could, however, be the case that this function wraps the panic! macro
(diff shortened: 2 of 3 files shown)


checker/src/callbacks.rs
```diff
@@ -183,6 +183,6 @@ impl rustc_driver::Callbacks for MiraiCallbacks {
             }
             info!("done with analysis");
         });
-        false // Do not proceed to code generation. MIRAI is only a checker.
+        true // Although MIRAI is only a checker we still need code generation for build scripts.
```
zlandau: Is there any danger here if people WANTED to use MIRAI and generate code for full projects? As in, before this change they could run MIRAI and then code generation as separate steps. Now that MIRAI will generate code, it would make more sense to do it in one go. Is that safe?

checker/src/visitors.rs
```diff
@@ -882,18 +882,27 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
...
-                        .already_report_errors_for_call_to
-                        .contains(&func_to_call)
-                    {
-                        self.emit_diagnostic_for_precondition(precondition, &message);
-                        self.already_report_errors_for_call_to
-                            .insert(func_to_call.clone());
+                    // false, so it may seem to make sense to just complain loudly here and not
+                    // promote the precondition, which can never be satisfied by the caller.
+                    // It could, however, be the case that this function wraps the panic! macro
+                    // and that the way to shut MIRAI up about it is to never call it.
```
zlandau: -that ?

checker/src/callbacks.rs
hermanventer: Sadly, it takes so long that nobody would want to do it. It is kind of safe in the sense that MIRAI does not make any changes to the MIR so the generated code is not affected, for better or worse, by the analysis itself. It is not quite safe in the sense that mirai-annotations behave differently when the code is compiled with MIRAI. Right now that only means that ([debug_]checked_)(verify|assume|precondition) become calls to dummy methods that do nothing at runtime.

In the future there may be some more differences as model state and trait contracts come into being.

checker/src/visitors.rs
hermanventer: Dunno, the paragraph parses for me.

PR #130, opened 2019-04-12
Tweak fixed point loop.

hermanventer:
## Description

The real change here is that the basic block list is now sorted so that for non loopy code a block is always analyzed after its predecessors. This speeds things up and makes widening better behaved (widening will come in a later PR).

However, while debugging this, I found an omission in the fixed point algorithm that stopped path conditions from propagating via blocks that do nothing of interest to MIRAI (other than propagating). Fixing this makes things more precise and making things more precise makes them more costly. Consequently there is now more aggressive time outs and another k-limit. Moreover a function whose analysis blows up, has been split in two.

Some of these regressions will be clawed back as I fix joining (in my next PR) and implement widening (in the PR after that).

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Also ran MIRAI on some of the crates it depends on.

### DIFF
--- a/checker/src/abstract_domains.rs
+++ b/checker/src/abstract_domains.rs
@@ -9,6 +9,7 @@ use crate::environment::Environment;
 use crate::expression::Expression::ConditionalExpression;
 use crate::expression::{Expression, ExpressionType};
 use crate::interval_domain::{self, IntervalDomain};
+use crate::k_limits;
 
 use rustc::ty::TyKind;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


checker/src/constant_domain.rs
```diff
@@ -387,23 +381,29 @@ impl ConstantDomain {
...
+                    ExpressionType::Isize => {
+                        isize::overflowing_mul(*val1 as isize, *val2 as isize).1
+                    }
                     ExpressionType::I128 => i128::overflowing_mul(*val1, *val2).1,
                     ExpressionType::I64 => i64::overflowing_mul(*val1 as i64, *val2 as i64).1,
                     ExpressionType::I32 => i32::overflowing_mul(*val1 as i32, *val2 as i32).1,
                     ExpressionType::I16 => i16::overflowing_mul(*val1 as i16, *val2 as i16).1,
                     ExpressionType::I8 => i8::overflowing_mul(*val1 as i8, *val2 as i8).1,
-                    _ => unreachable!(),
+                    _ => unreachable!("target_type {:?}", target_type),
```
zlandau: til

PR #131, opened 2019-04-12
Support unconditional join operations.

hermanventer:
## Description

The normal case for a join is when two branches that split on a condition come together again. The result of such a join can be expressed as a conditional expression.  Sometimes, however, two unconditional branches join together. This can happen for an infinite loop, where the dominator of the loop anchor unconditionally branches to the anchor and the backward branch also unconditionally branches to the anchor. While such loops can be expected to be rare in source form, they do show up during abstract interpretation because the loop exit condition may evaluate to false for the first few (precise) iterations of the loop and thus the backwards branch only becomes conditional once widening kicks in.

For such cases, we need a new kind of expression, which is what this PR introduces.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/expression.rs
+++ b/checker/src/expression.rs
@@ -127,6 +127,18 @@ pub enum Expression {
         right: Box<AbstractDomain>,
     },
 
+    /// Either left or right without a condition to tell them apart.
+    /// This can happen when there is loss of precision because of a loop fixed point computation.
+    /// For instance inside a loop body after the second fixed point iteration, a counter may have
+    /// either its initial value or the value computed in the first loop body iteration and we
+    /// don't have a way to tell which value it is.
+    Join {
+        // The value of the left operand.
+        left: Box<AbstractDomain>,
+        // The value of the right operand.
+        right: Box<AbstractDomain>,
+    },
+
     /// An expression that is true if left is less than or equal to right. <=
     LessOrEqual {
         // The value of the left operand.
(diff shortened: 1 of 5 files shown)


checker/src/expression.rs
```diff
@@ -127,6 +127,18 @@ pub enum Expression {
         right: Box<AbstractDomain>,
     },
 
+    /// Either left or right without a condition to tell them apart.
+    /// This can happen when there is loss of precision because of a loop fixed point computation.
+    /// For instance inside a loop body after the second fixed point iteration, a counter may have
+    /// either its initial value or the value computed in the first loop body iteration and we
+    /// don't have a way to tell which value it is.
+    Joined {
```
zlandau: I can't figure out if I like this name or not. It's a little bit confusing, because normally if two domains were joined you would just have..the joined domain. But I can't figure out a name that makes more sense to me. But either way, the description here explains it well, so meh.

hermanventer: If you look closely, you'll see that Expression is not a domain. Not just because it says so in the comments, but because, unlike ConstantDomain and IntervalDomain there is no Domain suffix.

The role of Expression is to keep track of that operations that resulted in the abstract value that refers to it. It therefore does make sense to me to name that Expression instance that results from a join operation after the operation.

I have to concede, however, that "Join" is probably more in keeping with the other enumeration labels than "Joined". So you actually do have a point here. I'll change it.

PR #132, opened 2019-04-12
Support widen operations.

hermanventer:
## Description

Currently widen always results in Top, which makes any variable written to inside a loop pretty much of a black box to MIRAI.

With this PR, the widen operation performs a normal join and then wraps the join in a Widen value, along with the path where the widened value will be stored (widen is always called when environments are joined/widened, so paths are available). Further widens just leave the value as is. 

When an interval is constructed from a widened join, the leftmost (i.e. earliest) value in the join tree is used to determine if the values in the widened set increase with every iteration, or decrease. If it is the one or the other, the upper (or lower) bound is removed from the returned interval.

Taken together with the path condition, it then becomes possible to constrain loop variables to ranges, which makes it possible to verify overflow checks and index out of bounds checks.

## Type of change

## How Has This Been Tested?
cargo test and validate.sh

### DIFF
--- a/checker/src/abstract_domains.rs
+++ b/checker/src/abstract_domains.rs
@@ -6,7 +6,7 @@
 use crate::abstract_value::{AbstractValue, Path};
 use crate::constant_domain::ConstantDomain;
 use crate::environment::Environment;
-use crate::expression::Expression::{ConditionalExpression, Join};
+use crate::expression::Expression::{ConditionalExpression, Join, Widen};
 use crate::expression::{Expression, ExpressionType};
 use crate::interval_domain::{self, IntervalDomain};
 use crate::k_limits;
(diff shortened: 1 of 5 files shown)


checker/src/abstract_domains.rs
```diff
@@ -901,7 +908,20 @@ impl AbstractDomain {
             Expression::Mul { left, right } => left.get_as_interval().mul(&right.get_as_interval()),
             Expression::Neg { operand } => operand.get_as_interval().neg(),
             Expression::Sub { left, right } => left.get_as_interval().sub(&right.get_as_interval()),
-            Expression::Variable { .. } => interval_domain::TOP,
+            Expression::Variable { .. } => interval_domain::BOTTOM,
+            Expression::Widened { operand, .. } => {
+                let interval = operand.get_as_interval();
+                if let Expression::Joined { left, .. } = &operand.expression {
+                    let left_interval = left.get_as_interval();
+                    if left_interval.lower_bound() == interval.lower_bound() {
```
zlandau: I may be missing something obvious, but in your description you say:

> When an interval is constructed from a widened join, the leftmost (i.e. earliest) value in the join tree is used to determine if the values in the widened set increase with every iteration, or decrease. If it is the one or the other, the upper (or lower) bound is removed from the returned interval.

But here you are seeing if the lower bounds (or upper bounds) are equal, not increasing/decreasing.

hermanventer: The interesting bound is the one that does not change because of the fixed point computation. That one we can keep, the other must be removed. But you are right to question this code. The version you reviewed did not do the right thing when an expression does not have an interval (the bottom case). I have rewritten it and have added comments that hopefully explain things better.

PR #135, opened 2019-04-17
Klimit config

### LINKED ISSUE #112: Make k-limits into command line options
## Issue

Currently a set of constants in checker/src/k_limits.rs provides limits on how much computation MIRAI will do in various places. With these limits in place, largish crates can be analyzed in a somewhat reasonable time frame. The limits do, however, impact on precision and may lead to false negatives. By making these into variables that can be set via command line options, it will become possible to schedule longer runs that may find more issues than more limited runs. The greater precision allowed by such runs can also be used to weed out false positives.

The recommended way to complete this task is to turn the constants in k_limits into the fields of a struct and to provide a constructor that sets the fields to default values, along with a command line processor that updates the values from options in the command line.

The limits struct should then be made part of the visitor state (checker/src/visitor.rs).

Ideally, there should also be a way to set these options to non default values for integration test cases. See also #52.

ryaneghrari:
## Description

Updated k-limits to command line options.  A default constructor will initialize an instance of KLimitsConfig with default values specified in k_limits.rs.  User can also initialize a custom config instance with the new constructor. 

Command line flags are as follows:
--max_analysis_time_for_body -> max_analysis_time_for_body
--max_outer_fixpoint_iterations -> max_outer_fixpoint_iterations
--max_path_length -> max_path_length
--max_inferred_preconditions -> max_inferred_preconditions

Fixes #112 (issue)

## Type of change

## Checklist:

- [x] Fork the repo and create your branch from `master`.
- [-] If you've changed APIs, update the documentation.

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -7,7 +7,6 @@ use crate::abstract_domains::{self, AbstractDomain};
 use crate::constant_domain::ConstantDomain;
 use crate::environment::Environment;
 use crate::expression::{Expression, ExpressionType};
-use crate::k_limits;
 
 use log::debug;
 use mirai_annotations::{assume, checked_assume};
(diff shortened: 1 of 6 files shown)


hermanventer:
Closing because of inactivity.

Issue #115, opened 2019-04-03
Implement custom serialization for AbstractValues


hermanventer:
## Issue

The automatically derived serializer/deserializer cannot deal with references, which makes it hard to define AbstractValue and related types in such a way that they can be cheaply shared.

As a first step to making AbstractValues serializable, just write explicit implications of the serializer methods for Summary and all types reachable from it.

hermanventer:
Serde can deal with Rc wrappers, so this is no longer needed.

PR #138, opened 2019-04-22
Clone less

hermanventer:
## Description

Expressions were passed by reference to the factory methods, which then had to clone them to incorporate them into the newly built expressions. This turns out to be really silly in practice because more often than not, a newly built expression is passed by reference to a factory, which then clones it, after which it is promptly freed when factory's caller exits.

In the long run, we should only refer to expressions via handles and never clone an expression, but for now this is an easy improvement.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Also checked proc-macro2 and its dependencies.

### DIFF
--- a/checker/src/environment.rs
+++ b/checker/src/environment.rs
@@ -58,10 +58,14 @@ impl Environment {
         if let Some((join_condition, true_path, false_path)) = self.try_to_split(&path) {
             // If path is an abstraction that can match more than one path, we need to do weak updates.
             let top = abstract_value::TOP;
-            let true_val = join_condition
-                .conditional_expression(&value, self.value_at(&true_path).unwrap_or(&top));
-            let false_val = join_condition
-                .conditional_expression(self.value_at(&false_path).unwrap_or(&top), &value);
+            let true_val = join_condition.clone().conditional_expression(
+                value.clone(),
+                self.value_at(&true_path).unwrap_or(&top).clone(),
+            );
+            let false_val = join_condition.conditional_expression(
+                self.value_at(&false_path).unwrap_or(&top).clone(),
+                value.clone(),
+            );
             self.update_value_at(true_path, true_val);
             self.update_value_at(false_path, false_val);
         }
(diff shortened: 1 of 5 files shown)


checker/src/environment.rs
```diff
@@ -58,10 +58,14 @@ impl Environment {
         if let Some((join_condition, true_path, false_path)) = self.try_to_split(&path) {
             // If path is an abstraction that can match more than one path, we need to do weak updates.
             let top = abstract_value::TOP;
-            let true_val = join_condition
-                .conditional_expression(&value, self.value_at(&true_path).unwrap_or(&top));
-            let false_val = join_condition
-                .conditional_expression(self.value_at(&false_path).unwrap_or(&top), &value);
+            let true_val = join_condition.clone().conditional_expression(
```
zlandau: You have to clone the join_condition to check it?

checker/src/environment.rs
```diff
@@ -162,17 +166,20 @@ impl Environment {
                 // abstract interpretation because the loop exit condition may evaluate to false
                 // for the first few iterations of the loop and thus the backwards branch only becomes
                 // conditional once widening kicks in.
-                x.join(y, p)
+                x.clone().join(y.clone(), p.clone())
```
zlandau: Same here, the x has to be cloned? Your diff to clone less has three clones on one line :)

checker/src/environment.rs
```diff
@@ -162,17 +166,20 @@ impl Environment {
...
         })
     }
 
     /// Returns an environment with a path for every entry in self and other and an associated
     /// value that is the widen of self.value_at(path) and other.value_at(path)
     pub fn widen(&self, other: &Environment, join_condition: &AbstractValue) -> Environment {
-        self.join_or_widen(other, join_condition, |x, y, _c, p| x.join(y, p).widen(p))
+        self.clone()
+            .join_or_widen(other, join_condition, |x, y, _c, p| {
+                x.clone().join(y.clone(), p.clone()).widen(p.clone())
```
zlandau: It gets better!

checker/src/environment.rs
hermanventer: I'm afraid so. join_condition is used again later in context where it is moved. I.e. I need two copies of it.

checker/src/environment.rs
hermanventer: Before this PR, join would do the cloning. Now the caller gets to decide. Sometimes, the caller knows that the value will not be reused. That causes fewer clones to happen at runtime.

checker/src/environment.rs
hermanventer: It does indeed. ;-)

hermanventer:
This PR only scratches the surface. MIRAI does way too much cloning and it is a real performance problem. Before this problem can be addressed, however, I need this task to get done: #115.

PR #140, opened 2019-04-23
Break up some big functions to reduce timeouts

hermanventer:
## Description

Functions with lots of loops take too long to analyze, so break up such functions in MIRAI to make it more amenable to analysis. While at it, also make the outer fixed point loop a bit more sensible and better documented. Also tweak things so that build scripts don't get analyzed (and don't interfere destructively with summary caching).

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -912,11 +912,12 @@ impl Path {
                 } else {
                     qualifier.replace_root(old_root, new_root)
                 };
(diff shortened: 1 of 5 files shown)


hermanventer:
It is nice to be able to make the sources friendlier and it that is what it takes, that is what I'll do. Sadly, this is not enough. The fixed points now converge, but the analysis still times out for pretty much all of these methods. Much more work remains to be done.

At any rate, keeping methods short is nothing but goodness and I should do it regardless of the fact that the analyzer chokes on them.

PR #142, opened 2019-04-24
Implement model fields.

hermanventer:
## Description

This PR adds the ability to attach specification only (model/ghost) fields to arbitrary values and to inspect them inside of verification conditions that are not checked at runtime (assume!, verify! and precondition!).

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
See the new test case, taint.rs, that shows how model fields can be used to check that tainted values are sanitized before being used in dangerous places.

### DIFF
--- a/checker/tests/run-pass/taint.rs
+++ b/checker/tests/run-pass/taint.rs
@@ -1,0 +1,46 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+#[macro_use]
+extern crate mirai_annotations;
+
+pub struct Foo {
+    bar: i64,
+}
+
+fn source() -> Foo {
+    let result = Foo { bar: 1 };
+    set_model_field!(&result, tainted, true);
+    result
+}
+
+pub fn sink(foo: Foo) {
+    // If foo has not been explicitly tainted by a source, it is not tainted.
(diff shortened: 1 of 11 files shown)


checker/tests/run-pass/taint.rs
```diff
@@ -0,0 +1,46 @@
...
+#[macro_use]
+extern crate mirai_annotations;
+
+pub struct Foo {
+    bar: i64,
+}
+
+fn source() -> Foo {
+    let result = Foo { bar: 1 };
+    set_model_field!(&result, tainted, true);
```
zlandau: One thing that threw me off here was when I saw this line, I was looking around for where 'tainted' was defined. I feel that:

  set_model_field(&result, "tainted", true);

would have been a bit more obvious to me. Alternatively, having some way to define the fields before they're used (like a struct), but that would feel weird here too. So I don't know. I still prefer having string keys, I think, but I don't feel strongly about it.

hermanventer: If you come from a language where, of course, meta programming is awkward and strings are the only way to identify things, then this is going to seem weird, at least at first. On the plus side, however, this seems less like meta programming (and that is the point) and it also makes it easy to think of tainted as a field, since the compiler insists on enforcing all of the identifier rules on this parameter. So, you can kind of imagine the macro resulting in `result.tainted = true`. Being able to declare such fields on the struct itself would be nice, and we did that in Spec#, but in practice that is going to be difficult for Rust.

PR #143, opened 2019-04-25
Fix precondition bugs

hermanventer:
## Description

Do not promote preconditions until the checking phase, otherwise they get duplicated.

Also simplify and clarify the precondition logic and fix a bug where an unknown entry condition would mask a (potential) precondition failure.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Modified a test case that is expected to report a false positive (which goes away with Z3 enabled).

### DIFF
--- a/checker/tests/run-pass/array_index.rs
+++ b/checker/tests/run-pass/array_index.rs
@@ -12,12 +12,11 @@ extern crate mirai_annotations;
 
 pub fn foo(arr: &mut [i32], i: usize) {
     arr[i] = 123; //~ possible array index out of bounds
-    // If we get here i is known to be within bounds, so no warning below.
-    bar(arr, i);
+    bar(arr, i); //~ possible array index out of bounds
 }
 
 fn bar(arr: &mut [i32], i: usize) {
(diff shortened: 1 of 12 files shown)


checker/tests/run-pass/array_index.rs
```diff
@@ -12,12 +12,11 @@ extern crate mirai_annotations;
 
 pub fn foo(arr: &mut [i32], i: usize) {
     arr[i] = 123; //~ possible array index out of bounds
-    // If we get here i is known to be within bounds, so no warning below.
-    bar(arr, i);
+    bar(arr, i); //~ possible array index out of bounds
```
zlandau: I take it that it's intentional that you get the warning twice now? (It makes sense to me, but given the comment being removed, I wanted to ask)

hermanventer: The removed comment is still true, but without Z3 (and the bug being fixed) the reasoning is not powerful enough to make use of the fact. Once Z3 gets merged in the comment will be back and the warning will be gone.

PR #144, opened 2019-04-26
Allow verify! and precondition! to have custom messages.

hermanventer:
## Description

Since preconditions get violated far away from their definitions, it can be very helpful to add descriptive text about what is expected. Likewise, any condition that get's checked at runtime can be improved with a custom message.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Updated existing tests.

### DIFF
--- a/checker/tests/run-pass/precondition.rs
+++ b/checker/tests/run-pass/precondition.rs
@@ -11,11 +11,11 @@ extern crate mirai_annotations;
 
 pub fn main() {
     let mut a = [1, 2];
-    foo(&mut a, 3); //~ unsatisfied precondition
+    foo(&mut a, 3); //~ unsatisfied precondition: i should be 0 or 1
 }
 
 fn foo(arr: &mut [i32; 2], i: usize) {
-    precondition!(i < 2); //~ related location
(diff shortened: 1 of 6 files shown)


checker/tests/run-pass/precondition.rs
```diff
@@ -11,11 +11,11 @@ extern crate mirai_annotations;
 
 pub fn main() {
     let mut a = [1, 2];
-    foo(&mut a, 3); //~ unsatisfied precondition
+    foo(&mut a, 3); //~ unsatisfied precondition: i should be 0 or 1
 }
 
 fn foo(arr: &mut [i32; 2], i: usize) {
-    precondition!(i < 2); //~ related location
+    precondition!(i < 2, "i should be 0 or 1"); //~ related location
```
zlandau: Using the first expression as the default message might be nice:

  precondition!(i < 2);

would print

  unsatisfied precondition: i < 2

just a thought.

hermanventer: It may well be, but doing so in the obvious way does not seem to work.

PR #147, opened 2019-04-26
Provide a way to return (and constrain) the result of a function spec.

hermanventer:
## Description

Function bodies that provide contracts for foreign/imported/abstract functions can be awkward to write in normal Rust if the function has a return result, since a specification may be abstract and the return result of a normal Rust function must be concrete. This PR adds a result!() macro that looks like a concrete value to the Rust compiler and that is understood by MIRAI as an abstract value that can be constrained with assertions.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
modified a test case

### DIFF
--- a/annotations/src/lib.rs
+++ b/annotations/src/lib.rs
@@ -532,6 +532,24 @@ macro_rules! get_model_field {
     };
 }
 
+/// Provides a way to refer to the result value of an abstract or contract function without
+/// specifying an actual value anywhere.
+/// This macro expands to unimplemented!() unless the program is compiled with MIRAI.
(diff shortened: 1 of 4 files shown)


annotations/src/lib.rs
```diff
@@ -532,6 +532,24 @@ macro_rules! get_model_field {
     };
 }
 
+/// Provides a way to refer to the result value of an abstract or contract function without
+/// specifying an actual value anywhere.
+/// This macro expands to unimplemented!() unless the program is compiled with MIRAI.
+/// It result should therefore not be assigned to a variable unless the assignment is contained
```
zlandau: It -> Its?

PR #150, opened 2019-05-03
Use name mangling to connect contracts to abstract/generic funcs

hermanventer:
## Description

Some traits have contracts that are difficult to describe accurately in the abstract. For example, the contract the generic iterator next function cannot say much that is definitive about the elements it returns. Certainly not enough for widening to come up with an expression that has a useful interval domain.

One way to deal with such situations is to have separate contracts for particular instantiations of the generic trait. For example, a separate contract for the iterator trait returned by RangeInclusive.into_iter function can result in a summary that allows the underlying counter increment to to become visible to the Abstract Interpreter.

This PR introduces the necessary mechanism to do this. The basic idea is to use the the type of the function constant used to call a generic trait function to discover the generic argument types and then to append the fully qualified signature of those types to the function summary key. If the summary cache has a summary for this key, that is used. If not, we fall back to using the generic key and its generic summary.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Beefed up the for_in.rs test case with foreign contracts for the core methods used under the cover of the the` for ordinal in 2..=n` construct.

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -8,7 +8,7 @@ use crate::environment::Environment;
 use crate::utils;
 
 use rustc::hir::def_id::DefId;
-use rustc::ty::TyCtxt;
+use rustc::ty::{Ty, TyCtxt};
 use serde::{Deserialize, Serialize};
 use sled::Db;
 use std::collections::{HashMap, HashSet};
@@ -186,6 +186,7 @@ pub struct PersistentSummaryCache<'a, 'tcx: 'a> {
     cache: HashMap<DefId, Summary>,
     dependencies: HashMap<DefId, Vec<DefId>>,
     key_cache: HashMap<DefId, String>,
+    argument_key_cache: HashMap<DefId, String>,
     type_context: &'a TyCtxt<'a, 'tcx, 'tcx>,
 }
 
@@ -201,6 +202,7 @@ impl<'a, 'tcx: 'a> PersistentSummaryCache<'a, 'tcx> {
                 .unwrap_or_else(|err| unreachable!(format!("{} ", err))),
             cache: HashMap::new(),
             key_cache: HashMap::new(),
+            argument_key_cache: HashMap::new(),
             dependencies: HashMap::new(),
             type_context,
         }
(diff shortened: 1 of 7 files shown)


checker/src/summaries.rs
```diff
@@ -240,26 +255,64 @@ impl<'a, 'tcx: 'a> PersistentSummaryCache<'a, 'tcx> {
...
+    /// Returns a summary from the persistent summary cache, preferentially using the concatenation
+    /// of persistent_key with arg_types_key as the cache key and falling back to just the
+    /// persistent_key if arg_types_key is None.
+    fn get_persistent_summary_using_arg_types_if_possible(
+        db: &Db,
+        persistent_key: &str,
+        arg_types_key: Option<&String>,
+    ) -> Summary {
+        if let Some(arg_types_key) = arg_types_key {
+            if !arg_types_key.is_empty() {
```
zlandau: FYI you can probably use slice patterns here (https://doc.rust-lang.org/edition-guide/rust-2018/slice-patterns.html).

checker/src/utils.rs
```diff
@@ -58,6 +59,139 @@ pub fn is_public(def_id: DefId, tcx: &TyCtxt<'_, '_, '_>) -> bool {
...
 }
 
+/// Returns a string that is a valid identifier, made up from the concatenation of
+/// the string representationss of the given list of argument types.
+pub fn argument_types_key_str<'tcx>(tcx: &TyCtxt<'_, '_, 'tcx>, ty: Ty<'tcx>) -> String {
+    let mut result = String::new();
+    let bound_sig = ty.fn_sig(*tcx);
+    let sig = bound_sig.skip_binder();
+    for arg_ty in sig.inputs().iter() {
+        result.push('_');
```
zlandau: _ isn't a valid part of the mangled type, I'm guessing?

checker/src/utils.rs
```diff
@@ -58,6 +59,139 @@ pub fn is_public(def_id: DefId, tcx: &TyCtxt<'_, '_, '_>) -> bool {
     }
 }
 
+/// Returns a string that is a valid identifier, made up from the concatenation of
+/// the string representationss of the given list of argument types.
+pub fn argument_types_key_str<'tcx>(tcx: &TyCtxt<'_, '_, 'tcx>, ty: Ty<'tcx>) -> String {
```
zlandau: There may be ambiguities here, right? Is that a potential problem?

checker/src/utils.rs
```diff
@@ -91,9 +225,15 @@ pub fn summary_key_str(tcx: &TyCtxt<'_, '_, '_>, def_id: DefId) -> String {
         } else {
             name.push('.');
         }
-        name.push_str(component.data.as_interned_str().as_str().get());
+        let component_name = component.data.as_interned_str().as_str();
+        let component_name = component_name.get();
+        if component_name == "{{impl}}" {
```
zlandau: Where does {{impl}} come from now? It looks like you removed the only instance of it? Or will this be used in future commits?

checker/src/summaries.rs
hermanventer: Probably, but it does not seem a clear improvement.

checker/src/utils.rs
hermanventer: It is a potential problem. Let me make it less likely by ensuring that it always violates a style rule.

checker/src/utils.rs
hermanventer: I'm none to sure what you are asking here.

checker/src/utils.rs
hermanventer: The Rust compiler sticks in the {{impl}} when a function is nested inside an impl block. I remove the braces because they are not valid in an identifier. The reason I do that is so that a foreign contract can define something using the mangled name.

checker/src/utils.rs
zlandau: Whoops, that was half a thought that I re-worded in argument_types_key_str. Ignore.

PR #152, opened 2019-05-06
Do not let promoted constants mess up the previous summary

hermanventer:
## Description

Looking into why the outer fixed point does not converge I found two minor problems that are fixed in this PR:
1) The default equality comparison for summaries compared the analysis time field, which clearly is neither deterministic nor relevant to the fixed point computation.
2) When a function body has embedded bodies for promoted constants, these bodies should not be summarized. Not only were they summarized, but the summaries were written to the cache, which made the before and after comparisons useless because the current (corrupted) version of the cache was used as the before version.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -41,7 +41,7 @@ use std::ops::Deref;
 ///    desirable to havoc all static variables every time such a function is called. Consequently
 ///    sound analysis is only possible one can assume that all such functions have been provided
 ///    with explicit contract functions.
-#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq, Hash)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq)]
 pub struct Summary {
     /// The number of seconds that the analyzer used to construct this summary.
     /// If it is >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY then this summary is not definitive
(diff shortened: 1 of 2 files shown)


checker/src/summaries.rs
```diff
@@ -94,6 +94,30 @@ pub struct Summary {
     pub unwind_side_effects: Vec<(Path, AbstractValue)>,
 }
 
+impl PartialEq for Summary {
```
zlandau: Not a fan of a single big return a == b && this == that and so on, eh?

checker/src/visitors.rs
```diff
@@ -535,6 +536,28 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
...
 
+    /// Computes a fixed point over the blocks of the MIR for a promoted constant block
+    fn visit_promoted_constants_block(&mut self, function_name: &str) {
+        let mut block_indices = self.get_sorted_block_indices();
+
+        let (mut in_state, mut out_state) =
+            <MirVisitor<'a, 'b, 'tcx, E>>::initialize_state_maps(&block_indices);
+
+        // The entry block has no predecessors and its initial state is the function parameters
+        // which we omit here so that we can lazily provision them with additional context.
```
zlandau: A bit of a nit, but since in general you make your explanations fairly descriptive: I found this a bit hard to parse, maybe just some parenthesis (assuming I even did manage to parse it correctly):

```
// The entry block has no predecessors and its initial state is the function parameters
// (which we omit here so that we can lazily provision them with additional context).
```

checker/src/summaries.rs
hermanventer: It really helps with debugging when you can set break points at particular parts of the expression.

PR #160, opened 2019-05-15
Provide a dedicated selector for Enum discriminants.

hermanventer:
## Description

The discriminant value of an Enum was treated bizarrely by associating it with the path of the entire enum. Made things more explicit, more like array lengths and less likely to cause subtle bugs.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/visitors.rs
+++ b/checker/src/visitors.rs
@@ -426,7 +426,7 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
             let i_state = if bb.index() == 0 {
                 first_state.clone()
             } else {
-                self.get_intial_state_from_predecessors(*bb, in_state, out_state, iteration_count)
+                self.get_initial_state_from_predecessors(*bb, in_state, out_state, iteration_count)
             };
             // Analyze the basic block
             in_state.insert(*bb, i_state.clone());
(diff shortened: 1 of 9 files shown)


checker/src/visitors.rs
```diff
@@ -632,7 +632,14 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
             "default visit_set_discriminant(place: {:?}, variant_index: {:?})",
             place, variant_index
         );
-        let target_path = self.visit_place(place);
+        let target_qualifier = self.visit_place(place);
+        let qualifier_length = target_qualifier.path_length();
+        assume!(qualifier_length < 1_000_000_000);
```
zlandau: I hope so :)

hermanventer:
This is a refactor that does not fix anything.

checker/src/visitors.rs
```diff
@@ -2198,18 +2226,24 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
...
             "default visit_discriminant(path: {:?}, place: {:?})",
             path, place
         );
-        let adtd_path = self.visit_place(place);
-        let adtd_type = self.get_place_type(place);
-        let adtd_value = self.lookup_path_and_refine_result(adtd_path, adtd_type);
-        self.current_environment.update_value_at(path, adtd_value);
+        let adt_path = self.visit_place(place);
+        let adt_path_length = adt_path.path_length();
+        assume!(adt_path_length < 1_000_000_000);
```
brenthaertlein: This magic number could be a well-defined constant?

hermanventer: It is pretty arbitrary. It just has to be big. The numeric literal is a good a name for it as any.

PR #163, opened 2019-05-17
Invalidate default summaries.

hermanventer:
## Description

If a function is referenced before it has been analyzed, the summary cache for the function will be populated with a default summary. Once the function is analyzed, the cache gets updated.

This is all and good, but it does not work for functions that live in a foreign_contracts module, since those functions have DefIds that are different from the real ones. For such functions we use spoofed persistent summary keys to populate the persistent cache with summaries that look like they come from the actual functions. Now, while the persistent (DB) cache's entries are updated when the contract functions get analyzed, the default summaries in the transient cache wont be updated because the update logic uses the DefIds of the contract functions, not the actual functions.

To work around this, any default summary that remains in the transient cache following an outer fixed point run are removed from the transient cache, so that the cache logic is forced to query the persistent (DB) cache with the spoofed key.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Modified an existing test to fail without this fix

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -41,12 +41,11 @@ use std::ops::Deref;
 ///    desirable to havoc all static variables every time such a function is called. Consequently
 ///    sound analysis is only possible one can assume that all such functions have been provided
 ///    with explicit contract functions.
-#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
 pub struct Summary {
-    /// The number of seconds that the analyzer used to construct this summary.
-    /// If it is >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY then this summary is not definitive
-    /// because the analyzer will omit it from the outer fixed point loop.
-    pub analysis_time_in_seconds: u64,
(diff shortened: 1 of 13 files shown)


checker/src/summaries.rs
```diff
@@ -41,12 +41,11 @@ use std::ops::Deref;
...
-#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
 pub struct Summary {
-    /// The number of seconds that the analyzer used to construct this summary.
-    /// If it is >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY then this summary is not definitive
-    /// because the analyzer will omit it from the outer fixed point loop.
-    pub analysis_time_in_seconds: u64,
+    /// If true this summary was computed. If false, it is a default summary.
+    /// Used to distinguish a computed empty summary from a default summary.
+    pub is_not_default: bool,
```
zlandau: I find 'is_default' a little more intuitive, otherwise you end up with '!is_not_default' and whatnot. But I guess if you manually avoid logic like that then it's ok. I'm guessing you did this because then a Summary::default() and whatnot will have the right value.

PR #164, opened 2019-05-20
DefId is not unique w.r.t. generic instantiations

hermanventer:
## Description

The DefId of a function constant is not a unique identifier if the constant is a method of a generic Trait instantiation. To ensure uniqueness, the type of the constant is now added to the key used to retrieve a function constant.

Also in this PR are two small tweaks to make type suffixes be valid identifiers.

## Type of change

## How Has This Been Tested?
cargo test; validate.sh
Added a new test case that fails without this fix.

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -8,7 +8,7 @@ use crate::environment::Environment;
 use crate::utils;
 
 use rustc::hir::def_id::DefId;
-use rustc::ty::{Ty, TyCtxt};
+use rustc::ty::TyCtxt;
 use serde::{Deserialize, Serialize};
 use sled::Db;
 use std::collections::{HashMap, HashSet};
@@ -183,9 +183,9 @@ fn extract_reachable_heap_allocations(
 pub struct PersistentSummaryCache<'a, 'tcx: 'a> {
     db: Db,
     cache: HashMap<DefId, Summary>,
+    typed_cache: HashMap<usize, Summary>,
     dependencies: HashMap<DefId, Vec<DefId>>,
     key_cache: HashMap<DefId, String>,
-    argument_key_cache: HashMap<DefId, String>,
     type_context: &'a TyCtxt<'a, 'tcx, 'tcx>,
 }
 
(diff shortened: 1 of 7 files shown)


checker/src/summaries.rs
```diff
@@ -210,6 +210,20 @@ impl<'a, 'tcx: 'a> PersistentSummaryCache<'a, 'tcx> {
     /// Any summaries that are still in a default state will be for functions in the foreign_contracts
     /// module and their definitions use DefIds that are different from the ids at the call sites.
     pub fn invalidate_default_summaries(&mut self) {
+        let keys_to_remove: Vec<usize> = self
```
zlandau: It'd be nice if the caches had an 'invalidate_default_summaries' method you could call, or at least a generic that implements the logic for both. I had to play 'spot the difference' again.

checker/src/summaries.rs
```diff
@@ -279,15 +280,50 @@ impl<'a, 'tcx: 'a> PersistentSummaryCache<'a, 'tcx> {
...
+    /// This matters if a useful function summary can only be written for specific generic instantiations.
+    pub fn get_summary_for_function_constant(
+        &mut self,
+        def_id: DefId,
+        function_id: usize,
+        persistent_key: &str,
+        arg_types_key: &str,
+        dependent_def_id: DefId,
+    ) -> &Summary {
+        let dependents = self.dependencies.entry(def_id).or_insert_with(Vec::new);
```
zlandau: Isn't  or_insert_with(Vec::new)  and or_default()  equivalent here?

checker/tests/run-pass/def_id_not_unique.rs
```diff
@@ -0,0 +1,44 @@
...
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// A test that call two distinct instantiations of the same generic functions in such a way
+// that both calls will use the same DefId, with the result that the second call will use
+// the wrong summary if using a cache that is keyed by DefId.
+
+// This example is highly contrived, but inspired by a larger way more subtle code fragment.
```
zlandau: supernit, but helps my brain parse this:

  // This example is highly contrived, but inspired by a larger, way more subtle, code fragment.

PR #176, opened 2019-06-03
Keep track of expression size

hermanventer:
## Description

There are some corner cases, in practice, where expressions can blow up exponentially in size. This leads to intolerable analysis times because of the consequent cost of expression refinement and serialization. To control this behavior, expressions now keep track of their size and join operations resort to widening if they would otherwise produce an expression that exceeds a k_limit. This is conservative in nature and trades precision for false positives.

Also in the PR a tweak to increase the maximum Path length to 30 because 10 is regularly exceeded and there does not seem to be a big cost to increasing it. Another tweak is to allow one more fixed point iteration before issuing an INFO log event.

With this change MIRAI can now run on itself again and finishes in less than a minute.

## Type of change

Performance trade-off.

## How Has This Been Tested?
cargo test; validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -37,6 +37,9 @@ pub struct AbstractValue {
     // It is used to refine the value with respect to path conditions and actual arguments.
     // It is also used to construct corresponding domain elements, when needed.
     pub expression: Expression,
+    // Keeps track of how large the expression is.
+    // When an expression gets too large it needs to get widened otherwise execution time diverges.
+    pub expression_size: u64,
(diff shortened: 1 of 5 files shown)


checker/src/abstract_value.rs
```diff
@@ -37,6 +37,9 @@ pub struct AbstractValue {
     // It is used to refine the value with respect to path conditions and actual arguments.
     // It is also used to construct corresponding domain elements, when needed.
     pub expression: Expression,
+    // Keeps track of how large the expression is.
+    // When an expression get's too large it needs to get widened otherwise execution time diverges.
```
zlandau: meganit: remove '

PR #178, opened 2019-06-04
Cache abstract values created from paths rooted in parameters

hermanventer:
## Description

Once an abstract (Variable) value is created to represent the value (or a part of the value) of a parameter, enter it into the environment so that subsequent references use the same value.

This makes the last fixed point divergence in MIRAI on MIRAI go away.

## Type of change

Code improvement.
Bug fix.

## How Has This Been Tested?
cargo test; ./validate.sh


hermanventer:
Not sure which expression sounds weirder to me. :-)

PR #181, opened 2019-06-05
Local variables that are not parameters need to get promoted.

hermanventer:
## Description

When a structured value is allocated on the stack as a local variable, initialized to an unknown value, and then copied into a field reachable from a mutable parameter, the unknown value leaks out to the caller in the form of a Variable expression with a path pointing to the callee's local variable.

This leads to naming conflicts and manifests as non linear behavior when analyzing MIRAI with MIRAI.

To fix this, an offset (that is large enough to eliminate naming conflicts in any realistic scenario) is added to the lead local variable ordinal, effectively promoting it into a local of the callee.

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh
new test case


checker/src/visitors.rs
```diff
@@ -886,6 +890,13 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
         from_hir_call: bool,
     ) {
         debug!("default visit_call(func: {:?}, args: {:?}, destination: {:?}, cleanup: {:?}, from_hir_call: {:?})", func, args, destination, cleanup, from_hir_call);
+        // This offset is used to distinguish any local variables that leak out from the called function
+        // from local variables of the callee function.
+        // This situation arises when a structured value stored in a local variable is assigned to
+        // a field reachable from a mutable parameter.
+        // We assume that no program that does not make MIRAI run out of memory will have more than
```
zlandau: Reminds me of this change to Python 3.7:

    More than 255 arguments can now be passed to a function, and a function can now have more than 255 parameters.

checker/src/abstract_value.rs
```diff
@@ -1326,92 +1326,100 @@ impl AbstractValueTrait for Rc<AbstractValue> {
     /// Returns a value that is simplified (refined) by replacing parameter values
     /// with their corresponding argument values. If no refinement is possible
     /// the result is simply a clone of this value.
-    fn refine_parameters(&self, arguments: &[(Rc<Path>, Rc<AbstractValue>)]) -> Rc<AbstractValue> {
+    fn refine_parameters(
+        &self,
+        arguments: &[(Rc<Path>, Rc<AbstractValue>)],
+        fresh: usize,
```
zlandau: The name 'fresh' meant very little to me until I saw the full name, fresh_variable_offset. It's not unless you find the comment in the visitor that it really makes sense. Not too big of a deal, but worth noting if you're concerned about the hackability of the codebase. Might warrant at least a comment to look at the description in the visitor file,.

checker/src/visitors.rs
hermanventer: We all assume something or other. 1 million is more than 640K, so sure it is going to be enough. :-)

PR #186, opened 2019-06-07
Add Vector model to foreign contracts + example program

darioncassel:
## Description

Added `std::vec::Vec` and `alloc::vec::Vec` models to `foreign_contracts.rs`. Added example that shows usage; `examples/shopping_cart.rs`. Updated `.gitignore` to not ignore new example.

Fixes: N/A

## Type of change

## How Has This Been Tested?

A new example called `shopping_cart.rs` has been written which tests the newly modeled vector functions.

## Checklist:

### DIFF
--- a/examples/foreign_contracts.rs
+++ b/examples/foreign_contracts.rs
@@ -324,5 +324,93 @@ pub mod std {
         }
     }
 
+    pub mod vec {
+        pub struct Vec<T> {
+            _phantom: std::marker::PhantomData<T>,
+            len: usize,
+        }
+        
+        impl<T> Vec<T> {
+            pub fn new() -> Vec<T> {
+                Vec {
+                    _phantom: std::marker::PhantomData,
+                    len: 0,
+                }
+            }
+
+            pub fn len(&self) -> usize {
+                self.len
+            }
+
+            pub fn push(&mut self, _value: T) {
+                precondition!(self.len < std::usize::MAX);
+                let old_len = self.len;
+                self.len += 1;
+                verify!(self.len == old_len + 1);
+            }
+
+            pub fn pop(&mut self) -> Option<T> {
+                if self.len == 0 {
+                    None
+                } else {
+                    let old_len = self.len;
+                    self.len -= 1;
+                    verify!(self.len == old_len - 1);
+                    result!()
+                }
+            }
+            
+            pub fn get(&self, index: usize) -> Option<&T> {
+                precondition!(index < self.len);
+                result!()
+            }
+        }
+    }
+
     pub mod result {}
+}
+
+pub mod alloc {
+    pub mod vec {
+        pub struct Vec<T> {
+            _phantom: std::marker::PhantomData<T>,
+            len: usize,
+        }
+        
+        impl<T> Vec<T> {
+            pub fn new() -> Vec<T> {
+                Vec {
+                    _phantom: std::marker::PhantomData,
+                    len: 0,
+                }
+            }
+
+            pub fn len(&self) -> usize {
+                self.len
+            }
+
+            pub fn push(&mut self, _value: T) {
+                precondition!(self.len < std::usize::MAX);
+                let old_len = self.len;
+                self.len += 1;
+                verify!(self.len == old_len + 1);
+            }
+
+            pub fn pop(&mut self) -> Option<T> {
+                if self.len == 0 {
+                    None
+                } else {
+                    let old_len = self.len;
+                    self.len -= 1;
+                    verify!(self.len == old_len - 1);
+                    result!()
+                }
+            }
+            
+            pub fn get(&self, index: usize) -> Option<&T> {
+                precondition!(index < self.len);
+                result!()
+            }
+        }
+    }
 }
--- a/examples/foreign_contracts.rs
+++ b/examples/foreign_contracts.rs
@@ -355,7 +355,7 @@ pub mod alloc {
             _phantom: std::marker::PhantomData<T>,
             len: usize,
         }
-        
+
         impl<T> Vec<T> {
             pub fn new() -> Vec<T> {
                 Vec {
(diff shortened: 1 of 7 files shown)


hermanventer:
This does look a bit more like a test case than an example proper.

darioncassel:
Updated with the shopping cart example expanded.

examples/foreign_contracts.rs
```diff
@@ -326,3 +348,43 @@ pub mod std {
...
+
+            pub fn len(&self) -> usize {
+                self.len
+            }
+
+            pub fn push(&mut self, _value: T) {
+                precondition!(self.len < std::usize::MAX);
+                let old_len = self.len;
+                self.len += 1;
+                verify!(self.len == old_len + 1);
```
hermanventer: There is no need to verify this. The previous line already has all the information we need from this contract.

examples/foreign_contracts.rs
```diff
@@ -326,3 +348,43 @@ pub mod std {
...
+            }
+
+            pub fn pop(&mut self) -> Option<T> {
+                if self.len == 0 {
+                    None
+                } else {
+                    let old_len = self.len;
+                    self.len -= 1;
+                    verify!(self.len == old_len - 1);
```
hermanventer: No need for this.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+        Cart {
+            items: Vec::new(),
+        }
+    }
+
+    fn add_item(&mut self, value: i32) {
+        precondition!(value > 0);
+        let old_len = self.items.len();
+        self.items.push(value);
+        verify!(self.items.len() == old_len + 1);
```
hermanventer: This just verifies that push has the obvious post condition. This is not a condition we would expect in real code.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
+#![feature(type_alias_enum_variants)]
+
+#[macro_use]
+extern crate mirai_annotations;
+
+pub mod foreign_contracts;
+
+struct Cart {
+    items: Vec<i32>
```
hermanventer: Vec<u32> seems more appropriate.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Cart {
+    fn new() -> Cart {
+        Cart {
+            items: Vec::new(),
+        }
+    }
+
+    fn add_item(&mut self, value: i32) {
+        precondition!(value > 0);
```
hermanventer: This is appropriate. The type system cannot express this, so the precondition adds value and gives MIRAI something to check at call sites.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+        let old_len = self.items.len();
+        self.items.push(value);
```
hermanventer: Although there is no need to say anything in the code here, interesting things are happening in MIRAI: push has a precondition that must be satisfied by this call site. To do so, MIRAI infers a precondition for add_item. The calls to add_item are then checked not just for the explicit precondition, but also for the inferred precondition.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+        verify!(self.items.len() == old_len + 1);
+    }
+
+    fn remove_item(&mut self) -> Option<i32> {
+        let old_len = self.items.len();
+        let res = self.items.pop();
+        verify!(self.items.len() == 0 || self.items.len() == old_len - 1);
```
hermanventer: No need to repeat the post condition of pop.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+            let x = self.items.get(i);
+            match x {
+                None => {},
+                Some(item) => {
+                    sum = sum + *item;
+                }
+            }
+            i += 1;
+        }
+        verify!(len <= 0 || sum > 0);
```
hermanventer: This is appropriate. It succinctly states a post condition for checkout. Making it explicit protects agains inadvertent relaxation of the post condition later on.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+        sum
+    }
+}
+
+fn main() {
+    let mut cart = Cart::new(); 
+
+    let old_num_items = cart.num_items();
+    cart.add_item(10);
+    verify!(cart.num_items() == old_num_items + 1);
```
hermanventer: Not useful.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+    let _ = cart.remove_item();
+    verify!(cart.num_items() == old_num_items);
```
hermanventer: Not useful.

examples/shopping_cart.rs
```diff
@@ -0,0 +1,72 @@
...
+        verify!(len <= 0 || sum > 0);
+        sum
+    }
+}
+
+fn main() {
+    let mut cart = Cart::new(); 
+
+    let old_num_items = cart.num_items();
+    cart.add_item(10);
```
hermanventer: Of interest here is that the implicit precondition for add_item is satisfied because MIRAI tracks the length value of cart.

examples/shopping_cart.rs
darioncassel: Can you clarify what you mean by this?

hermanventer: I meant u32

examples/shopping_cart.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    fn add_item(&mut self, value: u32) {
+        // We don't want items with zero
+        // value to be added to the cart
+        precondition!(value > 0);
+        self.items.push(value);
+    }
+
+    fn remove_item(&mut self) -> Option<u32> {
+        let res = self.items.pop();
```
hermanventer: The local variable is not necessary here.

PR #187, opened 2019-06-07
Always keep a hash with a path

hermanventer:
## Description

Embedded paths are looked up in the environment often enough for hashing to be significant. Overall the memory for computation trade-off seems like a wash or even slightly negative, but the trade-off does solve some time-out issues when running MIRAI on MIRAI, so overall it is a win.

## Type of change

- [x] Code improvement

## How Has This Been Tested?
cargo test; ./validate.sh

### DIFF
--- a/checker/src/path.rs
+++ b/checker/src/path.rs
@@ -22,40 +22,29 @@ use std::rc::Rc;
 /// recomputing it. The caching has a cost, so only use this in cases where it is highly likely
 /// that the path will be hashed more than once.
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, Ord, PartialOrd)]
-pub struct PathWithHash {
-    pub path: Rc<Path>,
+pub struct Path {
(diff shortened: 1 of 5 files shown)


checker/src/path.rs
```diff
@@ -65,7 +54,7 @@ impl From<Rc<Path>> for PathWithHash {
 /// When the actual expression is evaluated at runtime it will resolve to a particular memory
 /// location. During analysis it is used to keep track of state changes.
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub enum Path {
+pub enum PathEnum {
```
zlandau: nit: Kinda meh about the name PathEnum. PathVariant, PathKind? PathAccess? I guess it's all meh :)

hermanventer: Call an enum an enum.

PR #188, opened 2019-06-07
Simplify the path condition following an if-else

hermanventer:
## Description

The path condition following an if-else statement is of the form (x && y) || (x && !y), which simplifies to just x. When this pattern appears frequently enough, it is a form of exponential blow-up, so simplification really matters in this case.

With this change, MIRAI on MIRAI now completes in 21.64s without any time-outs. That is less time than it takes the Rustc compiler to do an optimized build.

Full disclosure: The outer fixed point loop is still restricted to just 2 iterations, which means that recursion is not properly analyzed. There is still much work to be done on that front and when that is fixed, the analysis time will no doubt be significantly more.

## Type of change

- [x] Performance bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
cargo test; ./validate.sh


hermanventer:
1) It was a minute or more.
2) Usually I change MIRAI to only analyze a single (problematic) function and do a profile on that. In this case I was pouring over the debug logs and noticed this pattern showing up a lot. I actually saw that a while back. At the time fixing it did not make a big difference and I did not want to complicate those diffs any further.

PR #190, opened 2019-06-17
More time-out checks and a k-limit check for expression refinement.

hermanventer:
## Description

Some funky functions take a long time to do things that were atomic w.r.t. time out checks. Made time-outs more granular. Also added a check that prevents very large expressions from being refined, thus trading precision for analysis speed.

## Type of change

- [x] Performance bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
cargo test; ./validate.sh

### DIFF
--- a/checker/src/visitors.rs
+++ b/checker/src/visitors.rs
@@ -1596,6 +1596,10 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                     break;
                 }
             }
+            let elapsed_time_in_seconds = self.start_instant.elapsed().as_secs();
+            if elapsed_time_in_seconds >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY {
+                return;
+            }
             self.current_environment.update_value_at(tpath, rvalue);
         }
     }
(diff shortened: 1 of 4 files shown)


checker/src/visitors.rs
```diff
@@ -1596,6 +1596,10 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                     break;
                 }
             }
+            let elapsed_time_in_seconds = self.start_instant.elapsed().as_secs();
+            if elapsed_time_in_seconds >= k_limits::MAX_ANALYSIS_TIME_FOR_BODY {
+                return;
```
zlandau: Do you keep stats on any of these types of timeouts? Isn't it hard to track down problems when anything can just silently give up?

checker/src/visitors.rs
```diff
@@ -1960,6 +1972,10 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
             .iter()
             .filter(|(p, _)| p.is_rooted_by(&rpath))
         {
+            let elapsed_time_in_seconds = self.start_instant.elapsed().as_secs();
```
zlandau: This is repeated a lot, isn't there somewhere you can put a reached_max_analysis_time? I guess it's cheating to put it into k_limits itself.

checker/src/visitors.rs
hermanventer: Time-outs surface as WARN level entries in log output. I am painfully aware of any time-out.

checker/src/visitors.rs
hermanventer: It is bothering me a bit as well, but not as much as other things. Perhaps I'll come back to this in the future.

PR #191, opened 2019-06-17
Add option to only analyze a single function

### LINKED ISSUE #113: Provide a command line option to specify analysis of a single function body
## Issue

When debugging MIRAI in a context where a small repro is not readily available, it is helpful to be able to run MIRAI in a mode where it will only analyze a specified method.

Typically, the summary cache from a full run will still be available, so analysis of a function body in this way can be done with much more context than is convenient to provide in a small test case.

To implement this feature, add a command line option (see also #112) and then add a check to the after_analysis function in checker/src/callbacks.rs that filters out any function that is not the one specified by the command line option.

darioncassel:
## Description

Add an environment variable `SINGLE_FUNC` that, if set to `fname`, restricts MIRAI to skip the analysis of all functions whose names don't match `fname`.

Fixes #113 

## Type of change

## How Has This Been Tested?

Testing has been completely locally. 

Linting via `clippy` fails due to the function `analyze_bodies` now having 8 arguments (over the max of 7 arguments). 

## Checklist:

- [x] If you've changed APIs, update the documentation.
- [ ] If you haven't already, complete your CLA here: <https://code.facebook.com/cla>

### DIFF
--- a/checker/src/callbacks.rs
+++ b/checker/src/callbacks.rs
@@ -30,7 +30,7 @@ pub struct MiraiCallbacks {
     /// True if this run is done via cargo test
     test_run: bool,
     /// If a function name is given, only analyze that function
-    analyze_single_func: Option<String>,
+    pub analyze_single_func: Option<String>,
 }
 
 /// Constructors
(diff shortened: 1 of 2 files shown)


darioncassel:
It seems that some refactoring will be necessary to reduce the number of arguments that `analyze_bodies` has. Perhaps `analyze_bodies` can have a reference to `self` and some of its arguments can be fields of the `MiraiCallbacks` struct?

checker/src/callbacks.rs
```diff
@@ -96,6 +100,11 @@ struct DefSets {
 }
 
 impl MiraiCallbacks {
+    /// Set MIRAI to analyze just a single function
+    pub fn set_analyze_single_func(&mut self, fname: Option<String>) {
```
hermanventer: A setter like this one does not add any value.

PR #196, opened 2019-06-19
Use the generic arguments to mangle function names.

### LINKED ISSUE #195: Name mangling fails on binding generic parameter via ::<>
## Issue

Name mangling fails when given a function bound with a generic parameter `T` via the syntax `::<>`.

Ex: `size_of::<u32>`

MIRAI will mangle the name of this function to be `size_of_`.

## Steps to Reproduce

Test case:
```
#![feature(type_alias_enum_variants)]
#[macro_use]
pub mod foreign_contracts;
use std::{ mem::{size_of} };

enum ConstantTest {}
impl ConstantTest { pub const OVERFLOW_U32: u32 = size_of::<u32>() as u32 * 2 + 1; }

fn main() { let _a = ConstantTest::OVERFLOW_U32; }
```

## Expected Behavior

Log output should be `[INFO  mirai::summaries] Summary store has no entry for core.mem.size_of_u32_usize`

## Actual Results

Log output is: `[INFO  mirai::summaries] Summary store has no entry for core.mem.size_of_`

## Environment

Rust version (rustc --version): `rustc 1.37.0-nightly (7cdaffd79 2019-06-05)`

hermanventer:
## Description

Function name mangling is currently done using parameter types. The reason we need it, however, is to distinguish between different generic instantiations and the current scheme fails to help when a generic function has no parameters.

Consequently, the mangling scheme now uses generic type arguments to do the mangling.

Fixes #195

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh
new test case

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 os: osx
 language: rust
 rust:
-- nightly-2019-06-06
+- nightly-2019-06-18
 
 before_install:
 - if [ ${TRAVIS_OS_NAME} == "osx" ]; then curl -L https://github.com/mozilla/grcov/releases/download/v0.4.3/grcov-osx-x86_64.tar.bz2 | tar jxf -; fi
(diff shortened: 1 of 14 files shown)


hermanventer:
Correct.

PR #214, opened 2019-06-27
Documentation

t-rasmud:
## Description

Adds link to Cargo build. Adds information on controlling MIRAI output verbosity.

## Type of change

- [x] Documentation change


documentation/Troubleshooting.md
```diff
@@ -0,0 +1,9 @@
+# Troubleshooting
+
+## Missing contracts
+Running MIRAI with log level set to `info` (`MIRAI_LOG=info`) gives information on missing contracts with the 
+message `Summary store has no entry for ...`.
+Add this missing contract to the file `standard_contracts/src/foreign_contracts.rs` and execute the script `rebuild_std.sh`
```
hermanventer: This should only be done if the missing contract is actually for a function in the Rust standard library. If it is for another crate, that crate should either be analyzed with MIRAI or foreign contracts should be added to to the crate.

In general, the documentation should be clearly bifurcated into documentation for MIRAI users and separate documentation for MIRAI developers. It may be best to even put them in separate directories.

PR #229, opened 2019-07-10
Use exit path condition of a function as its post condition.

### LINKED ISSUE #227: assume condition doesn't get change the path condition as expected.
## Issue

The condition in `assume` doesn't seem to be modifying the path condition as expected.

## Steps to Reproduce

Test case:
```
#[macro_use]

pub struct Block {
    round: u64,
}

pub fn round(bl: Block) -> u64{
    assume!(bl.round < std::u64::MAX - 1);
    bl.round
}

pub fn foo(bl: Block) {
    let ret = round(bl);
    verify!(ret < std::u64::MAX - 1);
}

```

## Expected Behavior

The `verify` condition is expected to be true.

## Actual Results

`thread 'rustc' panicked at 'Unexpected error: possible false verification condition`

## Environment

Rust version (rustc --version): `rustc 1.37.0-nightly (0beb2ba16 2019-07-02)`

hermanventer:
## Description

The path condition at the end of a function is effectively a post condition for the caller and it is now added to function summaries and transferred to call sites.

This additional precision has exacerbated exponential blow ups, even in the simple for_in test case. As part of looking into this, I've added some new simplification rules that help to keep expression size down. Ultimately, however, there needs to be a k-limit on expression size and the current check in refine_paths is too optimistic. I've lowered the limit and now enforce it on expression creation. If the limit is exceeded, TOP is returned instead.

The for_in test case remains broken and I'll keep looking into that, but it seems important to support post conditions sooner rather than later.

Sadly, the runtime for MIRAI on MIRAI has regressed quite a bit as well.

Fixes #227 

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -176,6 +176,9 @@ impl AbstractValue {
     /// Initializes the optional domains to None.
     #[logfn_inputs(TRACE)]
     pub fn make_from(expression: Expression, expression_size: u64) -> Rc<AbstractValue> {
+        if expression_size > k_limits::MAX_EXPRESSION_SIZE {
+            return Rc::new(TOP);
+        }
         Rc::new(AbstractValue {
             expression,
             expression_size,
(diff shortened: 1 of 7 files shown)


hermanventer:
It turns out that inferred post conditions are rather complex and often not all that useful. The additional complexity represents a significant extra computational cost and in some simple cases leads to exponential blow up that cannot be made to go away with simple tricks.

For the time being, I'm going to go with the requirement that post conditions must be supplied as explicit annotations.

PR #230, opened 2019-07-11
Distinguish between logical not and bitwise not.

### LINKED ISSUE #228: Lack of support for bitwise negation
## Issue

MIRAI seems to treat all expressions like `!e` as logical negation, while Rust does support bitwise negation operation. This makes Z3 crash due to typing errors.

## Steps to Reproduce

Run MIRAI for the utf8-range crate (v1.0.3) reproduces the crash. More specifically, the crash is triggered when analyzing the function [`Utf8Sequences::next`](https://github.com/BurntSushi/utf8-ranges/blob/eda2c387d1f4368933ebd53b89850bcdf11f1560/src/lib.rs#L334).

## Expected Behavior

The analysis should terminate without errors.

## Actual Results

An error message from Z3 before crash: `Error: Sort mismatch at argument #1 for function (declare-fun not (Bool) Bool) supplied sort is Int`

## Environment

Rust version 1.37.0-nightly (5f3656ce9 2019-06-11)

hermanventer:
## Description

MIRAI incorrectly assumed that !x is always a logical operation, whereas if x is an integer it amounts to a bitwise operation. This is now supported by inspecting the type of the operand and generating two different types of expression, depending on the type.

At the SMT solver level, it now need to understand bitwise not operations.

Fixes #228 

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh
New test case

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -199,6 +199,7 @@ pub trait AbstractValueTrait: Sized {
     fn as_bool_if_known(&self) -> Option<bool>;
     fn as_int_if_known(&self) -> Option<Rc<AbstractValue>>;
     fn bit_and(&self, other: Self) -> Self;
+    fn bit_not(&self, target_type: ExpressionType) -> Self;
     fn bit_or(&self, other: Self) -> Self;
     fn bit_xor(&self, other: Self) -> Self;
     fn cast(&self, target_type: ExpressionType) -> Self;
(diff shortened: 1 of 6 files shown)


uraj:
I tried this patch. Arithmetics over constants is still broken in some cases.
```rust
pub fn constants() {
    // this fails
    verify!(!0x11u8 == 0xeeu8);
    // this fails too
    verify!(!0x11u8 as u128 == 0xeeu128);
    // this passes
    verify!(!0x11u128 == 0xffffffffffffffffffffffffffffffeeu128);
}
```

hermanventer:
You are right that I forgot about the constant domain, but the problem was actually elsewhere: the result type of the bitwise not operation had to be obtained from MIR and not by inference on the operand type.

Thanks for noticing and providing the test case.

PR #240, opened 2019-07-20
Propagate TOP

hermanventer:
## Description

Add more checks to make sure that things like "TOP + x" result in TOP, which helps expressions not to grow unboundedly.

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -783,7 +783,7 @@ impl AbstractValueTrait for Rc<AbstractValue> {
         if self.is_bottom() {
             return other;
         }
-        // TOP union y is jut TOP
+        // TOP union y is just TOP
         if self.is_top() {
             return self.clone();
         }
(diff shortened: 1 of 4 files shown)


checker/src/abstract_value.rs
```diff
@@ -762,6 +783,10 @@ impl AbstractValueTrait for Rc<AbstractValue> {
         if self.is_bottom() {
             return other;
         }
+        // TOP union y is jut TOP
```
darioncassel: Typo `jut` -> `just`

validate.sh
```diff
@@ -11,7 +11,7 @@ cargo clean -p mirai
 # Run format checks
 cargo fmt --all
 # Run lint checks
-cargo audit
+#cargo audit
```
darioncassel: Why disable `cargo audit`?

hermanventer: There is a security alert that we cannot fix locally.

PR #246, opened 2019-07-23
Add annotations to get MIRAI on MIRAI clean of warnings.

hermanventer:
## Description

Add annotations and changes to address and remove all of the warnings that MIRAI currently gives on its own code.

## Type of change

## How Has This Been Tested?
cargo test; ./validate.sh

### DIFF
--- a/checker/src/summaries.rs
+++ b/checker/src/summaries.rs
@@ -11,6 +11,7 @@ use crate::path::{Path, PathEnum};
 use crate::utils;
 
 use log_derive::{logfn, logfn_inputs};
+use mirai_annotations::assume;
 use rustc::hir::def_id::DefId;
 use rustc::ty::TyCtxt;
 use rustc_errors::SourceMapper;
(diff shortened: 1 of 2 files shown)


checker/src/summaries.rs
```diff
@@ -598,10 +601,12 @@ impl<'a, 'tcx: 'a> PersistentSummaryCache<'a, 'tcx> {
     ) -> &Summary {
         let dependents = self.dependencies.entry(def_id).or_insert_with(Vec::new);
         if !dependents.contains(&dependent_def_id) {
+            assume!(dependents.len() < std::usize::MAX);
             dependents.push(dependent_def_id);
         }
         if self.typed_cache.contains_key(&function_id) {
-            self.typed_cache.get(&function_id).unwrap()
+            let result = self.typed_cache.get(&function_id);
+            result.expect("value disappeared from typed_cache")
```
darioncassel: Instead of this, perhaps we can just do `if let Some(result) = self.typed_cache.get(&function_id) { ... } else { ... }`?

checker/src/visitors.rs
```diff
@@ -314,6 +314,7 @@ impl<'a, 'b: 'a, 'tcx: 'b, E> MirVisitor<'a, 'b, 'tcx, E> {
                 already_added,
             );
         }
+        assume!(block_indices.len() < std::usize::MAX);
         block_indices.push(root_block);
```
darioncassel: Same as above comment about `assume!`

checker/src/summaries.rs
hermanventer: The point of this is to suggest that this failure can only happen if the code is changed incorrectly in the future, so the else case will have to be a failure as well. This formulation is more concise. Ideally, the unwrap should have been enough, but that requires a lot of standard contract writing, which I'm putting off for now.

PR #251, opened 2019-07-26
Adds standard contracts

t-rasmud:
## Description

Adds missing contracts for rust library functions.

## Type of change

## How Has This Been Tested?
`rebuild_std.sh`

### DIFF
--- a/standard_contracts/src/foreign_contracts.rs
+++ b/standard_contracts/src/foreign_contracts.rs
@@ -20,12 +20,22 @@ pub mod core {
             fn lt__ref_i32_ref_i32(x: &i32, y: &i32) -> bool {
                 (*x) < (*y)
             }
+            fn ge<T>(_a: &T, _b: &T) -> bool { result!() }
+            fn gt<T>(_a: &T, _b: &T) -> bool { result!() }
+            fn le<T>(_a: &T, _b: &T) -> bool { result!() }
+            fn lt<T>(_a: &T, _b: &T) -> bool { result!() }
         }
 
         pub trait Ord {
             fn cmp<T>(_a: &T, _b: &T) -> Ordering {
                 result!()
             }
+            fn min<T>(_a: &T, _b: &T) -> T { result!() }
+        }
+
+        pub trait PartialEq {
+            fn eq<T>(_a: &T, _b: &T) -> bool { result!() }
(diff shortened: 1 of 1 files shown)


standard_contracts/src/foreign_contracts.rs
```diff
@@ -20,12 +20,36 @@ pub mod core {
             fn lt__ref_i32_ref_i32(x: &i32, y: &i32) -> bool {
                 (*x) < (*y)
             }
+            fn ge<T>(_a: &T, _b: &T) -> bool {
```
hermanventer: This definition does not add much value. There are no preconditions and no useful data about the result. It may as well be the default summary. The only thing changed by this is that there is no longer a warning about a missing summary. Not having a warning in this case seems worse than having a warning.

standard_contracts/src/foreign_contracts.rs
```diff
@@ -77,12 +102,87 @@ pub mod core {
             ) -> Arguments<'a> {
                 result!()
             }
+
+            pub fn new_v1_formatted<'a>(
+                pieces: &'a [&'a str],
+                _args: &'a [ArgumentV1<'a>],
+                fmt: &'a [rt::v1::Argument],
+            ) -> Arguments<'a> {
+                precondition!(pieces.len() >= fmt.len());
```
hermanventer: There is not much value in this contract either, since the Rust compiler already checks that there is an argument for every piece of the format that requires an argument.

standard_contracts/src/foreign_contracts.rs
```diff
@@ -412,6 +512,20 @@ pub mod core {
...
                 pub range: Range_usize,
             }
+
+            pub struct Enumerate<I> {
+                _iter: I,
+                _count: usize,
+            }
+
+            impl<I> Enumerate<I> {
+                pub(super) fn new(iter: I) -> Enumerate<I> {
```
hermanventer: This summary adds value.

standard_contracts/src/foreign_contracts.rs
```diff
@@ -515,11 +630,47 @@ pub mod core {
...
+                        Self: Sized,
+                    {
+                        Enumerate::new(self)
+                    }
                 }
             }
         }
     }
 
+    pub mod time {
```
hermanventer: None of these summaries add value.

PR #269, opened 2019-08-20
Adds standard contract for max usize

t-rasmud:
## Description

Adds missing standard contract for `core::cmp::max__usize`.

## Type of change

## How Has This Been Tested?
`./rebuild_std.sh`


hermanventer:
Firstly, please always squash your commits. Secondly, ./rebuild_std.sh does not test anything. Ideally you should add a test case that only works if the new summary is in place. Lastly, while you are at it, why not add similar summaries for all of the other primitive comparable types?

PR #270, opened 2019-08-20
Standard contract for cmp::max

t-rasmud:
## Description

Adds missing contract for `core::cmp::max`

## Type of change

## How Has This Been Tested?
Test case added: `checker/tests/run-pass/cmp_max.rs`
testing command: `cargo test`


hermanventer:
Looks good. I'm looking forward to seeing it complete with summaries for all of the other primitive comparable types.

hermanventer:
Is there a reason you left out i128 and u128?

hermanventer:
The CI failure is because the binary version of the summary store was built with pagecache 0.19 whereas the checked-in lock file still refers to pagecache 0.18. You can fix this by running cargo update and then adding the resulting lock file to this PR.

PR #282, opened 2019-08-29
Try to enable CI on Windows

hermanventer:
## Description

Trying to see if MIRAI can be built on Windows with Travis


hermanventer:
Does not seem possible right now.

PR #330, opened 2019-10-31
Make unanalyzable functions "Angelic"

hermanventer:
## Description

A function that contains calls to other functions for which no body can be obtained and for which no summary has been provided via the standard summary store or the foreign_contracts mechanism cannot be accurately analyzed, leading to many false positives. Right now, the sheer number of messages that are meaningless and hard to make go away are a barrier to adoption.

Consequently "unanalyzable" functions are now marked as "angelic" (which is to say, we simply assume that they are correctly implemented). Any function that calls an angelic function itself becomes angelic, because the lack of post conditions from the angelic call makes the analysis of the caller inaccurate and noisy.

Needless to say, MIRAI now finds very few issues. With time this will improve as more functions without bodies are eliminated either by the top down analysis phase, or by explicit models.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
./validate.sh
export RUSTFLAGS="-Z always_encode_mir"
cargo clean
cargo build
find . -type f -name lib.rs -exec touch {} ;
RUSTC_WRAPPER=mirai cargo build

### DIFF
--- a/checker/src/visitors.rs
+++ b/checker/src/visitors.rs
@@ -73,6 +73,8 @@ pub struct MirVisitor<'analysis, 'compilation, 'tcx, E> {
 
     active_calls: Vec<hir::def_id::DefId>,
     already_reported_errors_for_call_to: HashSet<Rc<AbstractValue>>,
+    // True if the current function cannot be analyzed and hence is just assumed to be correct.
+    assume_function_is_angelic: bool,
     assume_preconditions_of_next_call: bool,
     async_fn_summary: Option<Summary>,
     check_for_errors: bool,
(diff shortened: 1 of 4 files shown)


checker/src/visitors.rs
```diff
@@ -145,6 +147,30 @@ impl<'analysis, 'compilation, 'tcx, E> Debug for MirVisitor<'analysis, 'compilat
     }
 }
 
+macro_rules! check_for_early_break {
```
wrwg: Add a short comment what this and the next one does?

hermanventer:
There is currently no scenario where a flag will be useful, nor do I foresee such a scenario arising in the near future.

wrwg:
> There is currently no scenario where a flag will be useful, nor do I foresee such a scenario arising in the near future.

Here is the scenario I have in mind: you really want to know whether there any functions with missing bodies anywhere transitively in a certain piece of code. With angelic on by default, you can't see it. Worth, you may turn a previously happily verified piece of code into something which is not longer verified accidentally by just changing a simple implementation detail somewhere deep down transitively.

Or do I misunderstand?

hermanventer:
You'll see such functions by turning on logging.

PR #337, opened 2019-11-06
Add simplifications

hermanventer:
## Description

The joining logic of the abstract interpreter, together with summary specialization result in many expressions that can be trivially simplified via pattern matching. This helps to reduce memory usage and prevent expressions from being prematurely widened, which speeds things up and helps precise reasoning when using Z3 to discharge proof obligations.

The simplification rules are necessarily ad-hoc and interact with each other in unpredictable ways, so expect lots of tweaking here.

Also, although the rules are mostly trivial, it is also easy to get them wrong, so please pay close attention to the comments that summarize the rules to see if a logic mistake has crept in.

I'm doing these additions now because they came up during debugging, when it really helps to have simple expressions.

I'll need them in future PRs.

## Type of change

## How Has This Been Tested?
./validate.sh
New test cases

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -377,6 +377,40 @@ impl AbstractValueTrait for Rc<AbstractValue> {
         } else if self.is_top() {
             other
         } else {
+            if let Expression::Or { left: x, right: y } = &self.expression {
+                // [(x || y) && x] -> x
+                // [(x || y) && y] -> y
+                if *x == other || *y == other {
+                    return other;
+                }
+                if let Expression::LogicalNot { operand } = &other.expression {
+                    // [(x || y) && (!x)] -> y
+                    if *x == *operand {
+                        return y.clone();
+                    }
+                    // [(x || y) && (!y)] -> x
+                    if *y == *operand {
+                        return x.clone();
+                    }
+                }
+            }
+            match (&self.expression, &other.expression) {
+                // [!x && !y] -> !(x || y)
+                (Expression::LogicalNot { operand: x }, Expression::LogicalNot { operand: y }) => {
+                    return x.or(y.clone()).logical_not();
+                }
+                // [!(x && y) && x] -> x
+                // [!(x && y) && y] -> y
+                (Expression::LogicalNot { operand }, _) => {
+                    if let Expression::And { left: x, right: y } = &operand.expression {
+                        if *x == other || *y == other {
+                            return other;
+                        }
+                    }
+                }
+                _ => (),
+            }
+
             // todo: #32 more simplifications
             let result = AbstractValue::make_binary(self.clone(), other.clone(), |left, right| {
                 Expression::And { left, right }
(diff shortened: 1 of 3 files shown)


checker/src/abstract_value.rs
```diff
@@ -377,6 +377,40 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
             other
         } else {
+            if let Expression::Or { left: x, right: y } = &self.expression {
+                // (x || y) && x == x
+                // (x || y) && y == y
+                if *x == other || *y == other {
+                    return other;
+                }
+                if let Expression::LogicalNot { operand } = &other.expression {
+                    //((x || y) && (!x)) = y
```
wrwg: nit: use == here and elsewhere for consistency. Also space to // and unnecessary parentheses.

checker/src/abstract_value.rs
```diff
@@ -377,6 +377,40 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                    if *x == *operand {
+                        return y.clone();
+                    }
+                    //((x || y) && (!y)) = x
+                    if *y == *operand {
+                        return x.clone();
+                    }
+                }
+            }
+            match (&self.expression, &other.expression) {
```
wrwg: For my taste, beginning with this match this fn has become too large. Split?

checker/src/abstract_value.rs
```diff
@@ -377,6 +377,40 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                // (!x && !y) == !(x || y)
```
wrwg: Are you attempting a CNF (instead of the usual DNF)? If so I'm not sure whether this is actually a simplification. I'm guessing the solver may convert this back to DNF.

checker/src/abstract_value.rs
```diff
@@ -377,6 +377,40 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                (Expression::LogicalNot { operand: x }, Expression::LogicalNot { operand: y }) => {
+                    return x.or(y.clone()).logical_not();
+                }
+                // !(x && y) && x
```
wrwg: Missing expected result. !(x && y) &&  x == y.

checker/src/abstract_value.rs
hermanventer: This is a heuristic that does three things:
1) Reduces the number of operators, so the expression is simpler in a strict sense.
2) Enables other simplifications to kick in.
2) Tends to transform expressions that are the ugly results of decompilation from MIR, back into expressions that look like the ones typically written by programmers.

checker/src/abstract_value.rs
hermanventer: I am trying to avoid overloading ==. The = here is a meta operator. I'll change it to -> so that it is less confusing.

wrwg: Just asking for consistency. Above you used ==

checker/src/abstract_value.rs
wrwg: I believe DNF is generally better for (3). For (2), pushing ! to the leaves generally increases potential for simplifications. But totally up to you.

hermanventer: My previous experience with simplification also points to DNF as the better option in general. This rule is helpful for debugging right now, but we might get back to it in the future.

checker/src/abstract_value.rs
hermanventer: Not right now. It is nice to have everything in one place and it is too early conceptually differentiate things into sub categories that logically belong together and can be read in isolation.

PR #338, opened 2019-11-07
Provide support for Option::and_then and similar functions

### LINKED ISSUE #311: assumed_postcondition not working correctly
## Issue

In attempting to run MIRAI against rust-smallvec, I ran into a case where the assumed_postcondition doesn't seem to propagate to the caller.

## Steps to Reproduce

Checkout https://github.com/metajack/badpostcond and run MIRAI on it.

## Expected Behavior

The `cap - len` should not be flagged, and this code should be MIRAI-clean.

## Actual Results

```
warning: possible attempt to subtract with overflow
  --> src/lib.rs:20:26
   |
20 |         let new_cap = if cap - len < additional {
   |                          ^^^^^^^^^

warning: possible unsatisfied postcondition
 --> src/lib.rs:27:9
  |
27|         postcondition!(new_cap >= cap);
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this warning originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

## Environment

1.40.0-nightly (8b0561d68 2019-09-30)

hermanventer:
## Description

Higher order functions like and_then are often parameterized by functions that must implement the FnOnce trait. The call to the argument function then goes via `FnOnce::call_once<T>` where T is a tuple type into which all of the arguments of the argument function must be packed. The implementation of call_once then invokes the argument function after checking some state (to ensure the call has not happened before) and unpacking the argument tuple via some tricky memory operations that are not easy to model in MIRAI.

Rather than proceeding all the way down the rabbit hole, this PR short cuts things by introducing a built-in model for `FnOnce::call_once<T>`.

Partially fixes #311

## Type of change

## How Has This Been Tested?
./validate.sh and building libra
New test cases added


checker/src/visitors.rs
```diff
@@ -1276,27 +1276,45 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+            destination,
+            &actual_args,
+            &function_summary,
+            func_to_call,
+        );
+        self.transfer_and_refine_cleanup_state(cleanup, &actual_args, &function_summary);
+    }
+
+    /// If we are checking for errors and have not assumed the preconditions of the called function
+    /// and we ware not in angelic mode and have not already reported an error for this call,
```
wrwg: s/ware/are

PR #339, opened 2019-11-08
An example of using contracts attributes.

wrwg:
## Description

This adds a new version of the shopping cart example (and removes the old one) which is richer and uses attributes from the contracts crate. The example is designed such contracts are executed (and tested) at runtime thus validating the soundness of the MIRAI assertions to the contracts crate.

Also attempted to verify this example using MIRAI and identified some gaps. See also TODO's in src/main.rs.

## Type of change

## How Has This Been Tested?

cargo test


.gitignore
```diff
@@ -2,6 +2,3 @@
 .vscode/
 .idea/
 target/
-examples/
-!examples/shopping_cart.rs
-!examples/taint/*.*
```
wrwg: Those exclusions actually don't work. From the doc: "It is not possible to re-include a file if a parent directory of that file is excluded." 

If we want to have a playground here it should be in a subdirectory or we . have a top-level 'playground'.

hermanventer: Lines 6 and 7 can go away.  Line 5 can stay. You can include something into this directory by using `git add -f xxx`

examples/shopping_cart/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+            && self.items.iter().map(|x| x.price).sum::<f64>() == self.total
+    }
+
+    #[post(ret.invariant())]
+    fn new() -> ShoppingCart {
+        return ShoppingCart { items: vec!(), total: 0.0 };
+    }
+}
+
+// Implements methods of the shopping cart. By attaching the the `#[invariant]` attribute
```
hermanventer: the the

examples/shopping_cart/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+// all methods in the impl which work on a self parameter get automatically injected
+// the invariant both as pre and post condition, in addition to what the state
+// individually.
+// TODO(wrwg): having both an invariant here and a post below is not supported
+// by MIRAI: "only one post condition is supported"
```
hermanventer: We should fix that.

examples/shopping_cart/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+        self.total = 0.0;
+        self.items.clear();
+        bill
+    }
+}
+
+// A main entry point which violates the invariant.
+pub fn main() {
+    let mut cart = ShoppingCart::new();
+    // The above should fail because pre-condition of Item::new is violated.
```
hermanventer: Below

examples/shopping_cart/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    total: f64,
+}
+
+impl ShoppingCart {
+    // Invariant for the shopping cart.
+    fn invariant(&self) -> bool {
+        // TODO(wrwg): this invariant is currently not handled by MIRAI:
+        // function DefId(2:1711 ~ core[5a18]::ops[0]::deref[0]::Deref[0]::deref[0])
+        // Comment the code out to see other issues with this code.
+        self.items.iter().all(|x| x.invariant())
```
hermanventer: It is going to be a while before MIRAI can deal with this kind of condition.

examples/shopping_cart/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+impl Item {
+    // Invariant for the Item struct. We define those invariants via functions
+    // so they can easily be referenced from attributes.
+    fn invariant(&self) -> bool {
+        self.name.len() > 0 && self.price > 0.0
+    }
+
+    // Creates a new Item, satisfying the invariant. Notice we can't use
+    // the `#[invariant]` attribute for new because this requires a `self` argument.
+    #[post(ret.invariant())]
```
hermanventer: The post condition implies a precondition, but I would prefer to see the precondition spelled out in a pre.

hermanventer:
You probably want to enable the setting in CLion that auto formats code when you save it.

.gitignore
wrwg: I did add everything manually with git add and then started to wonder why it was needed, and why the common -a workflow did not work. It took me a while to find the problem. I'm wondering why should the ./examples tree behave different than the others? Is this a well-known pattern I'm just not aware of?

Happy to change back, but don't understand the reasoning yet. As I said if you want to have a private playground in a repo tree, you could have a special subdir for that. I expect that examples grows over time, and will also serve as a set of integration tests.

hermanventer: Well, examples IS the current private playground in the repo tree.

examples/shopping_cart/src/main.rs
wrwg: Agreed, this is on top of the list.

examples/shopping_cart/src/main.rs
wrwg: Did it, but this is actually an interesting case which I expect to be quite common, and which needs a better solution. We essentially attaching a pre-condition to a pure data constructor which is identical with the invariant. If invariants become more complex, we don't want the author to need to duplicate the predicates. 

What I really would like to do here is:

    #[invariant(P)] 
    impl S {
      // Because this returns S, invariant should be applied to the result.
      // However, currently invariants are only applied to self arguments.
       fn new(params...) -> S { ... }
       ...
    }

I guess I open an issue for this on the contracts crate.

.gitignore
wrwg: Most projects on github and elsewhere use the 'examples' root directory of the repo for containing an authoritative set of usage examples of the sdk. Those are usually also part of the tests.

Because of this, it took me quite a while to consider looking in the .gitignore. I wasn't expecting a private playground in ./examples.

But never mind, we can address this later. Undone.

PR #341, opened 2019-11-09
Reenable multiple post conditions

wrwg:
## Description

This enables multiple post conditions by accumulating them via av.and() instead of issuing a warning for them.

## Type of change

## How Has This Been Tested?

./validate.sh (but no new tests)


wrwg:
This is a blind flight because I do not understand much of the code around. (E.g. is this accumulating for a given execution path or for the entire block?)

Also if there is a way how to test this semantically let me know (e.g. point to a similar test).

checker/tests/run-pass/post_conditions.rs
```diff
@@ -24,12 +24,12 @@ pub fn foo(bl: Block) {
...
-pub fn bad(c: bool) -> u64 {
+pub fn bar(c: bool) -> u64 {
     let result = result!();
     if c {
         assumed_postcondition!(result < 5);
     } else {
-        assumed_postcondition!(result > 10); //~ only one post condition is supported
+        assumed_postcondition!(result > 10);
     }
     result
```
hermanventer: It would be great if you can now add some calls to the function and verify that the results are as expected.

checker/src/visitors.rs
```diff
@@ -2084,23 +2084,24 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
-                if self.post_condition.is_some() {
-                    let span = self.current_span;
-                    let warning = self
-                        .session
-                        .struct_span_warn(span, "only one post condition is supported");
-                    self.emit_diagnostic(warning);
-                }
+                let extend_post_condition = |current: &Option<Rc<AbstractValue>>, cond| {
+                    if let Some(current_cond) = current {
+                        Some(current_cond.and(cond))
```
hermanventer: This and is not safe as is. In the test case below it will result in the a post condition that states `result < 5 && result > 10.` The desired post condition is instead `if c { result < 5 } else { result > 10 }`. Of course, doing the latter simplistically is also problematic because if might end up with things like `if c { result < 5 } else if c { result > 2 } else { result > 10 }.` Obviously we would want that to be `if c { result < 5 && result > 2 } else { result > 10 }`.

checker/tests/run-pass/post_conditions.rs
wrwg: That was one question I already asked: how to test this? Can you point to an example test which involves logic like this?

wrwg:
Closing this PR and opening a new one as it has drastically changed.

PR #343, opened 2019-11-11
Improving log messages for missing summaries

wrwg:
## Description

This improves the reporting (currently via logs) of functions which have been flagged as angelic. Specifically, the type instantiation is printed. Example:

    function ShoppingCart::invariant can't be analyzed because it calls function std::ops::Deref::deref which has no body and no summary (foreign fn argument key: __alloc_vec_Vec_shopping_cart_Item)

Because the full type information seems to be erased from FunctionReference, we use the existing summary_argument_key. This is also what a user would have to use if he wants to create a foreign contract.

Also fixes a warning about overflow in the shopping cart example.

## Type of change

## How Has This Been Tested?

./validate.sh and manual inspection of MIRAI run on shopping_cart example.


checker/src/visitors.rs
```diff
@@ -3,6 +3,27 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+use std::borrow::Borrow;
```
hermanventer: I like to keep these alphabetical and without random blank lines.

wrwg: I have not changed anything manually here. Some auto format in CLion did this (organize imports). The logic isn't random and the newlines have purpose (and are standard in many languages this way). Anyway, undone.

hermanventer: It seems like some great eminence decided that the std namespace needs special treatment. I can't say that I would like to adhere to this splendid wisdom.

wrwg: This goes back to java conventions: std libraries first, then 3rd party libraries, then your own project imports. It appears this has been taken over from CLion's roots (IntelliJ).

I love this feature. Automatic import management (adding, removing, ordering) is such a pleasure to work with in Java. I never look again at this part of the code (and it is actually folded away by default in intellij/android studio).

But when I have to look, if import lists become large, I also find it very useful for navigation.

hermanventer: Well, as long as Clippy and Format are happy, you are welcome put up a PR that sorts all of the import lists of all files in the repo.

PR #345, opened 2019-11-12
Provide built in summaries for transmute and deref.

### LINKED ISSUE #265: MIRAI reports a false warning with verify!
## Issue

MIRAI issues a false warning when a provably true condition is specified with `verify!`

## Steps to Reproduce
Test case:
```
use mirai_annotations::{verify};
use std::{
    sync::Arc,
};

pub type Round = u64;

pub struct Block
{
    round: Round,
}

impl Block
{
    pub fn round(& self) -> Round {
        let result = self.round;
        result
    }

    pub fn voting_rule(
        &mut self,
        proposed_block: Arc<Block>,
    )  {
        if proposed_block.round() <= self.round() {
            return;
        }
        verify!(proposed_block.round() > self.round());
    }
}
```

## Expected Behavior

No MIRAI warning.

## Actual Results

```
warning: possible false verification condition
  --> src/lib.rs:27:9
   |
27 |         verify!(proposed_block.round() > self.round());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this warning originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

## Environment

Rust version (rustc --version): `rustc 1.38.0-nightly (6e310f2ab 2019-07-07)`

hermanventer:
## Description

Transmute and deref are difficult to model via summaries written in Rust itself, so build in knowledge of them directly into the MIRAI, via the KnownFunctionNames mechanism.

Fixes #265

## Type of change

## How Has This Been Tested?
./validate.sh
New test case
Also ran MIRAI on libra and observed that many more functions are now analyzable.


checker/src/visitors.rs
```diff
@@ -1775,6 +1776,14 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
                 };
                 qualifier.unwrap_or_else(|| abstract_value::BOTTOM.into())
             }
+            KnownFunctionNames::StdOpsDeref => {
+                checked_assume!(args.len() == 1);
+                args[0].1.deref().clone().into()
```
wrwg: Could you add a short comment why this is supposed to work? It appears you are mapping this to identity? Or is this to be refined later?

If its identity, does this work for every deref, also user defined?

hermanventer: Good catch. deref is actually meant to be AbstractValue::dereference. The test case did not fail because there is some auto dereferencing going on elsewhere. I'm not sure I know how to quickly write a test case that will actually fail when mapping to identity, which is a bit sad.

PR #347, opened 2019-11-14
Enable contracts in integration tests

wrwg:
## Description

This enables use of contract annotations in integration tests. The integration test driver is extended to this end to deal with multiple --extern libraries.

A new test case is added based on contracts but this doesn't do much yet because of missing summaries.

## Type of change

## How Has This Been Tested?

./validate.sh


checker/tests/run-pass/contract_annotations.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn get(&self) -> i32 {
+        self.x
+    }
+    fn get_and_add(&mut self, x: i32) -> i32 {
+        let c = self.x;
+        // The below is currently needed because in MIRAI compilation mode, the runtime
+        // assertion generated by the matching #[pre] is dead code. Therefore the compiler
+        // does not know that this overflow is protected and issues a warning.
+        // TODO: find a way to fix this as it is a general problem
+        assert!(self.get() <= std::i32::MAX - x);
```
hermanventer: assert! is not part of the MIRAI contract language. You probably want to use checked_assume! here.

checker/tests/run-pass/contract_annotations.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        self.x = self.x + x;
+        return c;
+    }
+}
+
+fn use_trait() {
+    let mut a = MyAdder { x: 1 };
+    verify!(a.get() == 1);
```
hermanventer: If you want runtime checking here, use checked_verify!.

checker/tests/integration_tests.rs
```diff
@@ -31,50 +31,60 @@ use std::fs;
...
-        {
+        let file_name = entry.file_name().to_str().unwrap_or("");
+        if !file_name.starts_with(base_name) {
+            continue;
+        }
+        if entry.path().to_str().unwrap().contains(".dylib.dSYM/") {
             continue;
         }
-        if !file_name.to_str().unwrap_or("").ends_with(".rlib") {
+        if !file_name.ends_with(".rlib") && !file_name.ends_with(".dylib") {
```
hermanventer: On Windows dylib should be dll and on Linux it should be so.

checker/tests/run-pass/contract_annotations.rs
wrwg: Please look at the comment. The problem seems to be that the checked_assume (which the contracts already generate -- see pre conditions) is compiled into:

  if cfg!("mirai") { assume! ... } else { assert! ... }

When MIRAI is on the assert! will be dead code and rustc doesn't know that overflow has been checked for.

At least that is my explanation so far for this behavior.

hermanventer: When mirai has been specified then MIRAI == rustc and the assume! should be sufficient. Of course, my assumption has always been that when mirai is specified, the produced code is not actually run.

hermanventer: Also, checked_assume! is equivalent to assert! when mirai is not specified.

wrwg: I understand what checked_assume et. al do. I was assuming that actually the rustc checker generates this warning, because it flags the according method as angelic, but now I understand what is happening. Simplified example:

  #[pre(std::i32::MAX - x >= y)]
  fn(x,y) { x + y }

  ====v=====

  fn(x,y) {
    checked_precondition(std::i32::MAX - x >= y)]
    let f = || x + y;
    let ret = f();
    ...
    return ret;
} 

We don't seem to propagate any knowledge about constraints on x and y into the closure fn.

I changed the TODO in the code to document this problem.

hermanventer: The way things are supposed to work is that the closures get inferred pre-conditions and the invocation of the closure satisfies them with the specified precondition.

wrwg: Well that seems not to work. I added a minimal repro in a new test case `closure_precond_inference.rs` as part of this PR.

I think this PR here is good to go but waiting for your approval.

hermanventer: The PR has to pass the CI first.

wrwg: Passing now. I wish travis would send me an email...

PR #350, opened 2019-11-14
WIP: A test which demonstrates that trait contracts aren't working

wrwg:
## Description

This is a test case which shows fundamental issues with contracts on traits as generated by #[contract_trait]. For clarity, this doesn't use annotations but directly spells out the generated code.

Not sure whether we submit this as it is more for documentation.

## Type of change

## How Has This Been Tested?

./validate.sh


checker/tests/run-pass/trait_contracts.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Get the state of the adder for use in contracts. This is a common design pattern to
+    // access state of trait implementations on abstract trait level.
+    fn current(&self) -> i32;
+
+    // Original function decrement with pre/post embedded calling a newly introduced impl function.
+    // That's conceptually what the conract crate generates for pre/post on trait methods.
+    fn decrement(&mut self) {
+        // PROBLEM: This function is not analyzable because self.current() has no summary. The
+        // summary would look different for each implementer. At the callers side, we can't know
+        // it (by design of virtual calls).
```
hermanventer: If you want to write a contract for decrement, you better have a contract for current. If you can't have the latter, you are hosed.

In general, it seems like bad idea to write traits that must be implemented in ways that cannot be specified. Unfortunately, Rust has quite a few such traits.

In principle we can come up with some specification abstractions that capture the general notion of thing such as "addition" (for example by axiomatizing it). We should not, however, expect to be able to translate this into proof obligations that Z3 can deal with, any time soon. It will also dramatically increase proof times because we can no longer rely on built in theories. I don't expect to go down this path any time soon.

In practice, we'll have to explicitly specify the behavior of trait implementations in the areas where it cannot be inherited from the trait definitions.

wrwg: I tend to disagree. If you want to write contracts over polymorphic objects, you need to be able to specify with polymorphism, in this case uninterpreted observer functions.

The pattern of using observer functions on traits is ubiquitous, and found in many rust articles on the web. It is not a corner case. Whenever default methods are put into a trait and the trait is stateful, it is used. Pre/Post conditions on trait level can be seen as an instance of default methods.

Even though the observers may be uninterpreted, the usage in contracts puts constraints on them. This should be totally feasible and provable. It is a very different problem than when we have uninterpreted functions like Fn.call. 

Something which requires a longer brainstorm in person I guess.

hermanventer: MIRAI already support uninterpreted functions, but this is currently masked by angelic behavior. Once we have sufficient coverage of the standard libraries and successfully de-virtualize trait methods, we can dial back on the angelic behavior.

It is not clear to me, however, that uninterpreted functions will solve many problems. But if things can be done this way, I'm all for it.

wrwg: Well, without turning angelic skip off at least for functions as they are used *inside of constraints*, I don't think we can specify many useful features for traits. It will work only for purely functional traits for stateless objects.

Notice again that there is a difference in using an uninterpreted function in a constraint and in code. In straight forward execution, uninterpreted functions are relative useless. In constraints introduced by verification conditions, the bar is much lower. E.g.

     [post(self.state() > 1)]
     fn f() ...
     ...

     let mut s = S();
     s.f();
     verify(s.state() > 0); // should be no problem

hermanventer: Indeed.

hermanventer:
This PR is now subsumed by #368.

PR #351, opened 2019-11-15
Enabling multiple post conditions

wrwg:
## Description

This enables multiple post conditions provided they appear in the same execution path. Thus this allows something like:

    fn f(...) {
      ...
      post
      ...
      post
    }

... but disallows:

    fn f(...) {
      ...
      if ... {
        ...
        post
      } else {
        ...
        post
      }

The implementation uses the MIR control graph's `is_dominated_by` relation to determine whether a postcondition can be extended on the same execution path. Thus it is relative general.

The mirai_annotations for postconditions needed to be changed to make this work. It turns out that a mirai macro expands to something like `if true { ... } else { ... }`, which is not eliminated in the MIR, and not reflected in the dominator relation either. We now use conditional compilation via `#[cfg(mirai)]` for all post condition macros. We don't use it for other macros because using conditional compilation should be avoided as it does not type check the full source. (The `if false` will be certainly easily eliminated in backends.)

New tests have been added and warnings have been removed from existing tests.

There are a few followup issues coming out of this PR:

- We need to further restrict placement of post conditions. Essentially, any return point needs to be 
   checked for whether the post condition dominates it. Fortunately, as we now store the basic block
   of the last post condition, this can be easily done ("self.post_condition_block must dominate every 
   exit point").

- It appears that `result!` is not working as expected. See comments in test `post_conditions.rs`.

## Type of change

## How Has This Been Tested?

New test cases.

### DIFF
--- a/checker/src/visitors.rs
+++ b/checker/src/visitors.rs
@@ -87,6 +87,7 @@ pub struct MirVisitor<'analysis, 'compilation, 'tcx, E> {
     generic_arguments: Option<Vec<ExpressionType>>,
     heap_addresses: HashMap<mir::Location, Rc<AbstractValue>>,
     post_condition: Option<Rc<AbstractValue>>,
+    post_condition_block: Option<mir::BasicBlock>,
     preconditions: Vec<Precondition>,
     unwind_condition: Option<Rc<AbstractValue>>,
     unwind_environment: Environment,
@@ -111,6 +112,7 @@ struct SavedVisitorState<'analysis, 'tcx> {
     generic_arguments: Option<Vec<ExpressionType>>,
     heap_addresses: HashMap<mir::Location, Rc<AbstractValue>>,
     post_condition: Option<Rc<AbstractValue>>,
+    post_condition_block: Option<mir::BasicBlock>,
     preconditions: Vec<Precondition>,
     unwind_condition: Option<Rc<AbstractValue>>,
     unwind_environment: Environment,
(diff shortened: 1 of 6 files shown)


checker/src/visitors.rs
```diff
@@ -87,6 +87,7 @@ pub struct MirVisitor<'analysis, 'compilation, 'tcx, E> {
     generic_arguments: Option<Vec<ExpressionType>>,
     heap_addresses: HashMap<mir::Location, Rc<AbstractValue>>,
     post_condition: Option<Rc<AbstractValue>>,
+    post_condition_block: Option<mir::BasicBlock>,
```
hermanventer: It may be better to put the value and corresponding block into a tuple.

checker/src/visitors.rs
```diff
@@ -2284,6 +2286,32 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
         }
     }
 
+    fn try_extend_post_condition(&mut self, cond: &Rc<AbstractValue>) {
```
hermanventer: Always prefix a function with a comment that summarizes what it does, a bit more expansively than what the name suggests.

checker/src/visitors.rs
wrwg: I tried this in a first version, and concluded flattening it out into the global type leads to less boilerplate. We can look at it if we have more usage of the pair (Cond, BasicBlock).

checker/src/visitors.rs
wrwg: Done. Totally agree.

PR #353, opened 2019-11-16
Make known name recognition more efficient

hermanventer:
## Description

Currently every time a function reference is created, there is a linear search of a ever growing list of strings to see if the def id corresponds to a known function name. This is not scalable.

This PR introduces a cache that is keyed with def id and that lives for the during of an analysis run. Also, instead of constructing a summary cache key string and then comparing the string against known list (or using a second hash map for this), the rust compiler is asked for the def path (which it caches) and the def path is analyzed without additional memory allocations.

## Type of change

## How Has This Been Tested?
./validate.sh

### DIFF
--- a/checker/src/known_names.rs
+++ b/checker/src/known_names.rs
@@ -1,0 +1,230 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+use rustc::hir::def_id::DefId;
+use rustc::hir::map::{DefPathData, DisambiguatedDefPathData};
+use rustc::ty::TyCtxt;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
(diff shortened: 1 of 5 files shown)


checker/src/known_names.rs
```diff
@@ -0,0 +1,230 @@
...
+
+use rustc::hir::def_id::DefId;
+use rustc::hir::map::{DefPathData, DisambiguatedDefPathData};
+use rustc::ty::TyCtxt;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// Well known definitions (language provided items) that are treated in special ways.
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, Eq, PartialOrd, PartialEq, Hash, Ord)]
+pub enum KnownNames {
```
wrwg: Plural in a type name indicates a collection. I think this should be called KnowName. But not a blocker.

hermanventer: To be fair, an enumeration is a kind of collection. :-)

PR #357, opened 2019-11-19
Infer post conditions and avoid unnecessary join

hermanventer:
## Description

MIR return instructions are now guaranteed to be unique so there is no need to for the problematic join between the current state at the point of the return and the exit state (which is either empty or a join involving other returns). The join with the empty state makes for ugly expressions and perhaps causes fixed point convergence issues.

Now that things have been simplified, the path condition at the single return point seems like a good candidate for serving as an inferred post condition. It seems to work, but causes some test cases to behave differently. All but for_in.rs seems to behave as one should expect.

For_in probably breaks because the heuristic for deriving an interval domain from the widened join is brittle and probably fails to cope with the addition of the inferred post condition from the iterator next call. This is sufficiently complex to fix that it deserves a PR of its own and so this PR just changes to expected output to match the new less precise behavior.

The motivation for making this change at this time is because it came up while trying to debug failures that present themselves when tests are run without angelic behavior.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI on Libra


checker/src/visitors.rs
```diff
@@ -1370,6 +1368,11 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
             return;
         }
         let function_summary = self.get_function_summary(&func_to_call, &actual_args);
+        debug!("calling {:?}", func_to_call);
```
hermanventer: I find myself adding these statements every time I debug.

checker/src/visitors.rs
```diff
@@ -3147,7 +3150,7 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
     fn visit_discriminant(&mut self, path: Rc<Path>, place: &mir::Place<'tcx>) {
         let discriminant_path =
             Path::new_discriminant(self.visit_place(place), &self.current_environment);
-        let discriminant_type = ExpressionType::Isize;
```
hermanventer: This is somewhat arbitrary, but needs to be consistent.

PR #359, opened 2019-11-21
Use rustc resolve method to devirtualize

hermanventer:
## Description

The rust compiler has a convenient method for resolving a call to a trait method to the corresponding implementation method, based on the type information known at the call site. (It is a little bit strange that the MIR does not just refer to the implementation method in the first place, but perhaps it does when the optimizer is used.)

While this helps, it does not take care of all devirtualization, since the MIR obtained for a trait method that does have a body is not type specialized. Or at least, I don't yet know how to get hold of a type specialized body. In such cases the body might contain calls to virtual trait methods and the (generic) call site type information will not allow those call sites to resolve to the implementation method.

This PR also fixes a bug: The Deref and DerefMut traits were incorrectly implemented as well known functions by MIRAI. Fixing this caused some existing test cases to fall over because the now non trivial behavior of calls to deref and deref_mut uncovered some bugs in how paths are refined. The most painful such bug caused locals to leak into summaries. This is now fixed, or perhaps just less broken.

The for_in test case has now gone from broken to angelic because devirtualization causes some wonky predefined summaries to not come into the picture.

On the other hand, closure_precond_inference now works correctly.

Finally, now that less code becomes angelic, a few previous bugs with Libra crates have resurfaced. Those are now black listed again.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI on Libra


checker/src/path.rs
```diff
@@ -357,6 +360,35 @@ impl PathRefinement for Rc<Path> {
                 }
             }
             if let Some(val) = environment.value_at(&refined_qualifier) {
+                match &val.expression {
+                    Expression::Variable { path, .. } => {
+                        // if path is a deref when just drop it because it becomes implicit
```
wrwg: s/when/we

PR #364, opened 2019-11-27
More work on devirtualization

hermanventer:
## Description

Keep track of the actual type of the self parameter and use this to resolve virtual trait methods to their implementation methods in the actual type. So, if a call chain starts with a fully instantiated type, virtual methods that are not generic will resolve to actual methods. This resolves the issues that come up when dealing with the code rewrites done by the contracts crate.

This PR also adds a few minor fixes to resolve problems that manifest because of the more comprehensive analysis. Additionally, MIRAI no longer reports errors when the root of the analysis is not public.

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/visitors.rs
```diff
@@ -469,6 +486,69 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+                            } else {
+                                might_resolve = false;
+                            }
+                        }
+                    }
+                    self.tcx.mk_param_from_def(param_def)
+                });
+                if might_resolve {
+                    let param_env = self.tcx.param_env(self.def_id);
+                    debug!("devirt param_env {:?}", param_env);
```
wrwg: Want to keep those in?

hermanventer: Yes

Issue #52, opened 2019-01-25
Add a way to run an integration test with a non default compiler option


hermanventer:
## Issue

Compiler options can influence the MIR that gets generated. For example, the `Rvalue::Len` expression may be turned into a constant for `-Z mir-opt-level=n` where `n > 0`. Since the default level is 1, `cargo test` will not run in a mode where the code path that handles `Rvalue::Len` gets reached.

In order to test this case using `cargo test`, the code integration_test.rs needs to be enhanced to look for comments that specify which compiler options should be used to compile the test case.

## Steps to Reproduce

Add an info level log statement to the code that handles `Rvalue::Len` and observe that `MIRAI_LOG=info cargo test` produces no output from this log statement.

## Expected Behavior

When this issue is fixed, you can add the compiler option `-Z mir-opt-level=0` to the test case array_literal.rs and you should observe a log entry from your added log statement.
What you expected to happen. For example the error message you expected to see.

## Actual Results

No test coverage for `Rvalue::Len` when running `cargo test`.

PR #365, opened 2019-12-01
Enabling analysis of tests, plus turning angelic treatment off.

wrwg:
## Description

This adds general MIRAI flag handling, and adds new flags `--test_only` and `--flaky` which allow for analyzing test functions as well as turning off the transient propagation of angelic treatment in the presence of functions without a body. This flags are intended to be used together, but are technically independent.

`--test_only` will cause MIRAI to look for a test runner and only analyze those methods which are called by the test runner (i.e. those which have the `#[test]` attribute), plus methods in their
call graph. The implementation of this isn't as trivial as expected since the rust test runner uses a native method and builds data structures for the set of test methods, while the original `#[test]` is removed from the source by macro expansion. We use heuristics which result in a first implementation which may need further refinement.

One must use the `--test` (rustc) or `--tests` (cargo) option in addition to `--test_only` to get the tests analyzed.

`--flaky` will simply cause the analyzer to report with info log that it can't analyze a method reliably, but continue instead of bailing out. The flag is named such it can serve as a general disabler for false positive suppression in the future.

A new test has been added and the test infrastructure extended to be able to deal with those options (a marker in the source indicates for integration tests what to do).

This also relates to #52 as it adds a way to inject MIRAI_FLAGS into test sources for integration tests. We want to do something similar for RUSTC_FLAGS, which should be an easy extension of the existing logic.

## Type of change

## How Has This Been Tested?

New test, ./validate.sh

---
This change is [Reviewable](https://reviewable.io/reviews/facebookexperimental/mirai/365)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,29 @@ At this stage your code will be better documented and more readable. Perhaps you
 Set the verbosity level of output from MIRAI by setting the environment variable `MIRAI_LOG` to one of
 `info`, `warn`, `debug`, or `trace`.
 
-To go beyond this, super lint + better documentation phase, you'll need to be aware of a lot more things.
+You can also use the environment variable `MIRAI_FLAGS` to provide options to MIRAI. The value is a string which
+can contain any of the following flags:
+
+- `--test_only`: instructs MIRAI to analyze only test methods in your crate. You must also provide the `--tests`
+  option to the `cargo build` command to include those tests actually into your build.
+- `--diag=relaxed|strict`: configures level of diagnostics. With `relaxed` (the default) MIRAI
+   will not report errors which are potential 'false positives'. With `strict` it will
+   report such errors. 
+- `--single_func <name>`: the name of a specific function you want to analyze. 
+- `--`: any arguments after this marker are passed on to rustc.
+
+A more comprehensive command line interface for MIRAI is planned, but currently not implemented.
+
+## Using MIRAI together with the contracts crate
+
+Preliminary support for MIRAI is available in the [contracts crate](https://gitlab.com/karroffel/contracts). There
+is currently no official release containing this support on crates.io, so you must directly refer to the gitlab
+repo using a dependency like below in your Cargo.toml:
+```toml
+contracts = { git = "https://gitlab.com/karroffel/contracts.git", branch = "master", features = [ "mirai_assertions" ]}
+```
+
+See the shopping cart example for usage.
 
 ## Developing MIRAI
 See the [developer guide](https://github.com/facebookexperimental/MIRAI/blob/master/documentation//DeveloperGuide.md)
--- a/checker/src/callbacks.rs
+++ b/checker/src/callbacks.rs
@@ -7,6 +7,7 @@
 use crate::constant_domain::ConstantValueCache;
 use crate::expected_errors;
 use crate::known_names::KnownNamesCache;
+use crate::options::Options;
 use crate::summaries::PersistentSummaryCache;
 use crate::utils;
 use crate::visitors::{MirVisitor, MirVisitorCrateContext};
@@ -14,7 +15,10 @@ use crate::z3_solver::Z3Solver;
 
 use log::info;
 use log_derive::{logfn, logfn_inputs};
-use rustc::hir::def_id::DefId;
+use rustc::hir::def_id::{DefId, LOCAL_CRATE};
+use rustc::mir;
+use rustc::session::config::ErrorOutputType;
+use rustc::session::early_error;
 use rustc::ty::TyCtxt;
 use rustc_driver::Compilation;
 use rustc_interface::interface;
@@ -24,39 +28,38 @@ use std::fmt::{Debug, Formatter, Result};
 use std::iter::FromIterator;
 use std::ops::Deref;
 use std::path::PathBuf;
-use std::rc::Rc;
 use syntax::errors::{Diagnostic, DiagnosticBuilder};
 use tempdir::TempDir;
 
 /// Private state used to implement the callbacks.
 pub struct MiraiCallbacks {
+    /// Options provided to the analysis.
+    options: Options,
     /// The relative path of the file being compiled.
     file_name: String,
     /// A path to the directory where analysis output, such as the summary cache, should be stored.
     output_directory: PathBuf,
     /// True if this run is done via cargo test
     test_run: bool,
-    /// If a function name is given, only analyze that function
-    pub analyze_single_func: Option<String>,
 }
 
 /// Constructors
 impl MiraiCallbacks {
-    pub fn new() -> MiraiCallbacks {
+    pub fn new(options: Options) -> MiraiCallbacks {
         MiraiCallbacks {
+            options,
             file_name: String::new(),
             output_directory: PathBuf::default(),
             test_run: false,
-            analyze_single_func: None,
         }
     }
 
-    pub fn test_runner() -> MiraiCallbacks {
+    pub fn test_runner(options: Options) -> MiraiCallbacks {
         MiraiCallbacks {
+            options,
             file_name: String::new(),
             output_directory: PathBuf::default(),
             test_run: true,
-            analyze_single_func: None,
         }
     }
 }
@@ -69,7 +72,7 @@ impl Debug for MiraiCallbacks {
 
 impl Default for MiraiCallbacks {
     fn default() -> Self {
-        Self::new()
+        Self::new(Options::default())
     }
 }
 
@@ -121,11 +124,13 @@ impl rustc_driver::Callbacks for MiraiCallbacks {
 }
 
 struct AnalysisInfo<'compilation, 'tcx> {
+    options: &'compilation Options,
     persistent_summary_cache: PersistentSummaryCache<'tcx>,
     constant_value_cache: ConstantValueCache<'tcx>,
     diagnostics_for: HashMap<DefId, Vec<DiagnosticBuilder<'compilation>>>,
     known_names_cache: KnownNamesCache,
-    analyze_single_func: Option<String>,
+    // Functions to include in analysis. If None, all functions are included.
+    function_whitelist: Option<Vec<String>>,
 }
 
 impl MiraiCallbacks {
@@ -184,17 +189,19 @@ impl MiraiCallbacks {
             "storing summaries for {} at {}/.summary_store.sled",
             self.file_name, summary_store_path
         );
+        let options = std::mem::replace(&mut self.options, Options::default());
         let persistent_summary_cache = PersistentSummaryCache::new(tcx, summary_store_path);
         let defs = Vec::from_iter(tcx.body_owners());
         let constant_value_cache = ConstantValueCache::default();
         let diagnostics_for: HashMap<DefId, Vec<DiagnosticBuilder<'_>>> = HashMap::new();
         let known_names_cache = KnownNamesCache::create_cache_from_language_items();
         let mut analysis_info = AnalysisInfo {
+            options: &options,
             persistent_summary_cache,
             constant_value_cache,
             diagnostics_for,
             known_names_cache,
-            analyze_single_func: self.analyze_single_func.to_owned(),
+            function_whitelist: None,
         };
         Self::analyze_bodies(compiler, tcx, &defs, &mut analysis_info);
         if !self.emit_or_check_diagnostics(&mut analysis_info.diagnostics_for) {
@@ -250,6 +257,28 @@ impl MiraiCallbacks {
         defs: &[DefId],
         analysis_info: &mut AnalysisInfo<'analysis, 'tcx>,
     ) {
+        // Determine the functions we want to analyze.
+        if let Some(fname) = &analysis_info.options.single_func {
+            analysis_info.function_whitelist = Some(vec![fname.clone()]);
+        } else if analysis_info.options.test_only {
+            // Extract test functions from the main test runner.
+            if let Some((entry_def_id, _)) = tcx.entry_fn(LOCAL_CRATE) {
+                let fns = Self::extract_test_fns(tcx, entry_def_id);
+                if fns.is_empty() {
+                    early_error(
+                        ErrorOutputType::default(),
+                        "Could not extract any tests from main entry point",
+                    );
+                }
+                analysis_info.function_whitelist = Some(fns);
+            } else {
+                early_error(
+                    ErrorOutputType::default(),
+                    "Did not find main entry point to identify tests",
+                );
+            }
+        }
+
         // Analyze all public methods
         for def_id in defs.iter() {
             if !utils::is_public(*def_id, tcx) {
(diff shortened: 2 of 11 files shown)


checker/src/callbacks.rs
```diff
@@ -338,7 +387,74 @@ impl MiraiCallbacks {
...
+            let tests = Self::extract_test_fns(tcx, def_id);
+            info!("found test fns: {:?}", tests);
+            return Some(tests);
+        }
+        None
+    }
+
+    /// Extract test functions from the promoted constants of a test runner main function.
+    /// This uses a heuristic matching the name out of a formatted string of a statement
+    /// in the body. This method might be bridle and we might want to change it if this runs
```
hermanventer: might be bridle -> seems brittle

checker/src/callbacks.rs
```diff
@@ -338,7 +387,74 @@ impl MiraiCallbacks {
...
+        if !name.contains(".main") {
+            // Rust generates a main function for #[test] targets which can be named
+            // main, main_1, etc. depending on an existing regular main function. We bail
+            // out if we don't find at least `.main` in the name.
+            return None;
+        }
+        // We determine whether this is a test runner by matching test::test_main_static in the
+        // string representation of the body statements.
+        info!("checking for test runner in {}", name);
+        let s = Self::get_statements_as_string(tcx.optimized_mir(def_id));
```
hermanventer: This makes me queasy.

checker/src/main.rs
```diff
@@ -40,12 +41,15 @@ fn main() {
         env_logger::init_from_env(e);
     }
 
-    // Configure the analysis to just run on a single function if set
-    let analyze_single_func = if env::var("SINGLE_FUNC").is_ok() {
-        env::var("SINGLE_FUNC").ok()
+    // Configure the analysis target
+    let target = if env::var("SINGLE_FUNC").is_ok() {
```
hermanventer: I should not have let SINGLE_FUNC go by me, but now seems a time to call a halt: We should not keep inventing environmental variable names that are not prefixed with MIRAI and should fix this one too. Generic names that might clash with names used by other tools create an adoption barrier for MIRAI.

checker/src/visitors.rs
```diff
@@ -468,18 +471,24 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
             {
                 return devirtualized_summary;
             }
-            self.assume_function_is_angelic = true;
+            self.assume_function_is_angelic = self.assume_function_is_angelic || !self.no_angelic;
```
hermanventer: It is not immediately obvious to me why the first disjunction term is needed.

checker/tests/integration_tests.rs
```diff
@@ -149,6 +151,20 @@ fn invoke_driver(
     sys_root: String,
     extern_deps: Vec<(&str, String)>,
 ) -> usize {
+    // Configure callbacks. This reads the source to check for some configuration
```
hermanventer: This is relevant to issue #52, please remember to add a comment to it when this is ready to merge.

checker/tests/integration_tests.rs
```diff
@@ -149,6 +151,20 @@ fn invoke_driver(
...
+    // options.
+    let mut call_backs = callbacks::MiraiCallbacks::test_runner();
+    {
+        let file_content = read_to_string(&Path::new(&file_name)).unwrap();
+        call_backs.target = if file_content.contains("//=TEST_ONLY") {
```
hermanventer: Re #52, a more general syntax might be appropriate.

checker/tests/run-pass/test_source.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[macro_use]
+extern crate mirai_annotations;
+
+//=TEST_ONLY
+//=NO_ANGELIC
+
+#[test]
+fn some_test() {
+    verify!(1 == 1);
```
hermanventer: Test methods are likely to use assert!, or in the case where the programmer is MIRAI aware, checked_verify! After all, we expect that these tests will have interesting runtime behavior.

checker/tests/run-pass/test_source.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+    let d: &dyn Dynamic = &S {} as &dyn Dynamic; // forget type info of S
+    verify!(d.f(1) == 3); // in angelic mode, this statement would disable verification for this
+                          // function and we would not see the message below. With no_angelic,
+                          // we do not see an error here (as d.f is uninterpreted), but we still
+                          // see the below.
+    verify!(3 == 4); //~provably false verification condition
+}
+
+#[allow(dead_code)]
```
hermanventer: Or just make the function public, in which case the property you are testing will actually be tested. As it is now, there won't an error message anyway for non public functions.

checker/src/callbacks.rs
```diff
@@ -36,8 +49,10 @@ pub struct MiraiCallbacks {
     output_directory: PathBuf,
     /// True if this run is done via cargo test
     test_run: bool,
-    /// If a function name is given, only analyze that function
-    pub analyze_single_func: Option<String>,
+    /// The target of the analysis.
+    pub target: AnalysisTarget,
+    /// True if angelic treatment of methods shall be turned off.
+    pub no_angelic: bool,
```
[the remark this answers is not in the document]
wrwg: I'm using now --flaky and explain it as a general switch which disables measurements for suppressing false positives. The semantics would be implementation dependent.

checker/src/callbacks.rs
wrwg: Replaced.

checker/src/main.rs
wrwg: Introduced now an argument parser based on the clap crate.

README.md
```diff
@@ -35,7 +35,28 @@ At this stage your code will be better documented and more readable. Perhaps you
 Set the verbosity level of output from MIRAI by setting the environment variable `MIRAI_LOG` to one of
 `info`, `warn`, `debug`, or `trace`.
 
-To go beyond this, super lint + better documentation phase, you'll need to be aware of a lot more things.
+You can also use the environment variable `MIRAI_FLAGS` to provide options to MIRAI. The value is a string which
+can contain any of the following flags:
+
+- `--test_only`: instructs MIRAI to analyze only test methods in your crate. You must also provide the `--tests`
+  option to the `cargo build` command to include those tests actually into your build.
+- `--flaky`: instructs MIRAI to disable measurements for omitting false positives. The output may be a lot
```
hermanventer: flaky is generally a pejorative term and we should avoid it. How about --verbose?

wrwg:
This has been updated, and is ready for another round.

README.md
wrwg: Flaky is a standard notion in test engineering over the industry. It was also introduced right away in the test talk in bootcamp.

wrwg: I'm also not happy with --flaky, but --verbose seems to be even more off, because it has a particular meaning in CLIs which is not exactly what we are doing here. I was also thinking of --aggressive, --imprecise,  --loose, --unfettered, --unfiltered. Any preferences or additional ideas?

hermanventer: Let's call it "--sound" then. It may not be completely sound now, but we can make that more true as time goes on.

"disable measurements" -> "disable measures".

"may include false negatives" -> "may include true positives"

checker/src/callbacks.rs
```diff
@@ -121,11 +122,12 @@ impl rustc_driver::Callbacks for MiraiCallbacks {
 }
 
 struct AnalysisInfo<'compilation, 'tcx> {
+    options: &'compilation Options,
     persistent_summary_cache: PersistentSummaryCache<'tcx>,
     constant_value_cache: ConstantValueCache<'tcx>,
     diagnostics_for: HashMap<DefId, Vec<DiagnosticBuilder<'compilation>>>,
     known_names_cache: KnownNamesCache,
-    analyze_single_func: Option<String>,
+    fn_whitelist: Option<Vec<String>>,
```
hermanventer: rather spell out fn

checker/src/callbacks.rs
```diff
@@ -250,6 +254,23 @@ impl MiraiCallbacks {
...
+        } else if analysis_info.options.test_only {
+            // Extract test functions from the main test runner.
+            if let Some((entry_def_id, _)) = tcx.entry_fn(LOCAL_CRATE) {
+                let fns = Self::extract_test_fns(tcx, entry_def_id);
+                if fns.is_empty() {
+                    warn!("Could not extract any tests from main entry point");
+                }
+                analysis_info.fn_whitelist = Some(fns);
+            } else {
+                warn!("Did not find any suitable main entry point to identify tests");
```
hermanventer: This raises the useless question of what makes an entry point suitable.

"any suitable" -> "a"

checker/src/callbacks.rs
hermanventer: Why is this an Option?

checker/src/callbacks.rs
```diff
@@ -250,6 +254,23 @@ impl MiraiCallbacks {
...
     ) {
+        // Determine the functions we want to analyze.
+        if let Some(fname) = &analysis_info.options.single_func {
+            analysis_info.fn_whitelist = Some(vec![fname.clone()]);
+        } else if analysis_info.options.test_only {
+            // Extract test functions from the main test runner.
+            if let Some((entry_def_id, _)) = tcx.entry_fn(LOCAL_CRATE) {
+                let fns = Self::extract_test_fns(tcx, entry_def_id);
+                if fns.is_empty() {
+                    warn!("Could not extract any tests from main entry point");
```
hermanventer: Rather than producing log output, provide user feedback in the same way as the rust compiler, via the early_error function.

See rust/src/librustdoc/lib.rs for an example.

checker/src/callbacks.rs
```diff
@@ -372,4 +387,51 @@ impl MiraiCallbacks {
...
+    /// method (via const test in the example). However, the constant slice in main will be promoted
+    /// into a constant initializer function in MIR, so we need to look there.
+    /// We therefore iterate overall promoted functions belonging to the main function, looking for
+    /// a statement like `_n = const <test name>` which loads the constant for a given test.
+    ///
+    /// This method is indeed not very stable and may break on changes to the compilation
+    /// scheme or the test framework.
+    fn extract_test_fns(tcx: TyCtxt<'_>, def_id: DefId) -> Vec<String> {
+        let mut result = vec![];
+        for body in tcx.promoted_mir(def_id).iter() {
```
hermanventer: This should be optimized_mir

checker/src/main.rs
```diff
@@ -40,12 +42,17 @@ fn main() {
...
+    // Get MIRAI options from environment variable. For now, we do not pass them
+    // on the command line because `RUSTC_WRAPPER=mirai cargo build` does not accept any
+    // options and arguments which it does not known about (even though we could remove them in
+    // this code here and not pass on to rustc). Once we have cargo integration
+    // of MIRAI we may want to change this.
+    let mut options = Options::default();
+    let mut rustc_args = vec![]; // any arguments after `--` for rustc
+    if env::var("MIRAI_FLAGS").is_ok() {
+        rustc_args = options.parse_from_str(&env::var("MIRAI_FLAGS").ok().unwrap());
+    }
```
hermanventer:
```
if let Ok(flags) = env::var("MIRAI_FLAGS") {
    rustc_args = options.parse_from_str(flags);
}
```

checker/src/options.rs
```diff
@@ -0,0 +1,65 @@
...
+// LICENSE file in the root directory of this source tree.
+
+use clap::{App, AppSettings, Arg};
+use itertools::Itertools;
+use shellwords;
+
+/// Creates the clap::App metadata for argument parsing.
+fn make_options_parser<'a>() -> App<'a, 'a> {
+    // We could put this into lazy_static! with a Mutex around, but we really do not expect
+    // to construct this more then once per regular program run.
```
hermanventer: I expect to see a comment when something is made static, since statics are usually evil, but there is no need to justify not making something static.

checker/src/options.rs
```diff
@@ -0,0 +1,65 @@
...
+        .long("flaky")
+        .takes_value(false)
+        .help("Report diagnostics even with incomplete information.")
+        .long_help("This removes measurements to suppress flaky diagnostics. The effect is implementation dependent."))
+}
+
+/// Represents options passed to MIRAI.
+#[derive(Debug, Default)]
+pub struct Options {
+    pub single_func: Option<String>,
```
hermanventer: May as well add the "tion".

checker/src/options.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub test_only: bool,
+    pub flaky: bool,
+}
+
+impl Options {
+    /// Parse options from an argument string. The argument string will be split using unix
+    /// shell escaping rules. Any content beyond the leftmost `--` token will be returned
+    /// (excluding this token).
+    pub fn parse_from_str(&mut self, s: &str) -> Vec<String> {
+        self.parse(&shellwords::split(s).unwrap())
```
hermanventer: Only use unwrap when MIRAI can prove it will never fail. If this is an assumption, use expect.

In this case I suspect things can fail because of bad user input. In such cases I want to see a proper user error message.

checker/src/visitors.rs
```diff
@@ -465,18 +469,25 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
             {
                 return devirtualized_summary;
             }
-            self.assume_function_is_angelic = true;
+            // Assume the function as angelic unless we have set the flaky option.
+            self.assume_function_is_angelic = !self.options.flaky;
```
hermanventer: Hah. Not flaky == Angelic. That is a low bar. :-)

Sound works. If you must, "demonic".

checker/src/visitors.rs
```diff
@@ -465,18 +469,25 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
             let argument_type_hint = if let Some(func) = func_ref {
                 format!(" (foreign fn argument key: {})", func.argument_type_key)
             } else {
                 "".to_string()
             };
+            let treatment_hint = if !self.options.flaky {
```
hermanventer: May as well use "self.assume_function_is_angelic" here.

checker/src/visitors.rs
```diff
@@ -465,18 +469,25 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+                "Excluding this function and its transitive callers from analysis.".to_string()
```
hermanventer: I suspect that converting this to a String (builder) is not necessary.

checker/src/options.rs
wrwg: I wish the code would have systematic error handling and user messaging setup. Instead, we are using unstable rustc APIs for which documentation is hard to find, and writing into logs which should arguably user messages. There should be a diag module somewhere where all the needed functionality is tugged away in an easy discoverable way, then it would be easier to follow rules.

In any case, I'm going to call early_error here now.

checker/src/callbacks.rs
wrwg: Settled offline for --diag=strict. Documented in the code why its an option, and renamed stuff.

checker/src/callbacks.rs
wrwg: promoted_mir is something else AFAICT. It delivers all the promoted function's mir as an iterator, not the mir of this function. I have not seen anything like promoted_optimized_mir.

checker/src/options.rs
wrwg: Its rather common to use this abbreviation. The 'fn' as well. I keep it (but have renamed fn).

checker/src/callbacks.rs
hermanventer: This is probably fine. It returns blocks of MIR that created promoted constants. A promoted constant is something that needs an address. Apparently this is where the test functions live.

PR #366, opened 2019-12-04
Propagate generic arguments to callees

hermanventer:
## Description

Propagate generic arguments from caller to callee by specializing the type parameters of the callee with the corresponding actual type arguments of the caller. (This is something the Rust compiler really ought to do when you ask for the MIR corresponding to a def_id that is type specialized.)

## Type of change

## How Has This Been Tested?
./validate.sh
New test case


checker/src/visitors.rs
```diff
@@ -497,44 +493,19 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
-                            } else {
-                                might_resolve = false;
-                            }
-                        }
-                    }
-                    self.tcx.mk_param_from_def(param_def)
-                });
-                if might_resolve {
+            if let Some(gen_args) = generic_args {
+                if utils::are_concrete(&gen_args) {
```
hermanventer: This is a bit sad, but without it the rust compiler panics in all sorts of ways that are hard to explain. I spent two days trying to figure out how to do better, but this is what I can get to work right now.

checker/src/visitors.rs
```diff
@@ -497,44 +493,19 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+                    debug!("devirt gen_args {:?}", gen_args);
                     let param_env = self.tcx.param_env(self.def_id);
                     debug!("devirt param_env {:?}", param_env);
-                    debug!("devirt def_id {:?}", def_id);
-                    debug!("devirt substs {:?}", substs);
                     if let Some(instance) =
-                        rustc::ty::Instance::resolve(self.tcx, param_env, def_id, substs)
+                        rustc::ty::Instance::resolve(self.tcx, param_env, def_id, gen_args)
```
hermanventer: This call panics if it does not like the arguments for some reason or another. Much nicer would be to return a Result.

checker/src/visitors.rs
```diff
@@ -1776,11 +1748,13 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
         match &value.expression {
             Expression::CompileTimeConstant(ConstantDomain::Str(s)) => s.clone(),
             _ => {
-                let error = self.session.struct_span_err(
-                    self.current_span,
-                    "this argument should be a string literal, do not call this function directly",
-                );
-                self.emit_diagnostic(error);
+                if self.check_for_errors {
```
hermanventer: Not related to the this PR, but I noticed it because of some bewildering panics in the rust compiler.

checker/tests/run-pass/promoted_slice.rs
```diff
@@ -17,94 +17,96 @@ pub fn promote_u8(value: &[u8]) -> u8 {
     const TAG: &[u8] = &[1, 2, 3];
     match value {
         TAG => 10,
-        _ => 20
+        _ => 20,
```
hermanventer: Nothing to see here, just updating the formatting of a test that failed at some point.

PR #368, opened 2019-12-05
Propagate the actual self type from caller to callee

hermanventer:
## Description

Adding another part to the puzzle of devirtualization: Calling a function that is generic in terms of its self parameter from a function that is not.

This has been painful to get to this point, mainly because rustc::ty::Instance::resolve panics in situations that are really hard to predict and control. For now, the painful work around is to provide built-in summaries for situations that fail in practice. 

Unfortunately, just not doing this is not acceptable either.

## Type of change

## How Has This Been Tested?
./validate.sh
new test case


checker/tests/run-pass/trait_contracts.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// This test case demonstrates that de-virtualization for contracts on traits as they
+// are generated by the contracts crate are infeasible, and that the contracts crate
```
wrwg: Should update this comment.

PR #369, opened 2019-12-05
Add more contracts and update Libra black list

hermanventer:
## Description

This breaks out the foreign contract additions that came up while trying to all of the test suite to run without resorting to viral angelic behavior.

Of course, most of these contracts are angelic themselves and will need to get fleshed out or moved into MIRAI itself, but that will have to wait a bit longer.

With the recent changes, many more Libra crates cause the analysis to fail, so I'm also expanding the black list.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


rebuild_std.sh
```diff
@@ -18,7 +18,7 @@ cargo install --debug --path ./checker
 
 # build the mirai-standard-contracts crate
 touch standard_contracts/src/lib.rs
-RUSTC_WRAPPER=mirai RUST_BACKTRACE=1 MIRAI_LOG=warn MIRAI_START_FRESH=true MIRAI_SHARE_PERSISTENT_STORE=true cargo build --lib -p mirai-standard-contracts
+RUSTFLAGS="-Z force-overflow-checks=off" RUSTC_WRAPPER=mirai RUST_BACKTRACE=1 MIRAI_LOG=warn MIRAI_START_FRESH=true MIRAI_SHARE_PERSISTENT_STORE=true cargo build --lib -p mirai-standard-contracts
```
wrwg: Yes, I think this is required. Most programs don't care about checking for overflow (and in reality, it most of the time doesn't matter).

hermanventer: In this case it is only to make it easier to use Rust to write summaries for functions that do no check for overflow.

PR #373, opened 2019-12-14
Refactor call related arguments into a helper struct

hermanventer:
## Description

This is a largely mechanical refactoring that collects together a group of parameters that are used as a group and passed as a group, into an actual structure.

Additionally, another parameter that could be part of a structure is bundled into that.

The consolidation of the call related arguments caused some changes in behavior as well because the somewhat ad hoc use of def id, function reference and function value has become a bit less ad hoc. This probably fixes some things, but it also causes a resolution error to show up in the trait contract test.

Fixing that is proving to be a challenge that will require a lot of additional code. This PR is already large and very hard to understand and I don't want to add to it, so I've disabled the test in order to move on. 

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/visitors.rs
```diff
@@ -126,6 +128,35 @@ struct SavedVisitorState<'analysis, 'tcx> {
     unwind_environment: Environment,
 }
 
+#[derive(Clone, Debug)]
+struct CallInfo<'callinfo, 'tcx> {
```
wrwg: +1 for trying to reduce arguments and organizing things in structs.

checker/src/visitors.rs
```diff
@@ -1508,22 +1497,44 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+                return def_id;
+            }
+            TyKind::Ref(_, ty, _) => {
+                if let TyKind::Closure(def_id, _) = ty.kind {
+                    return def_id;
+                }
+            }
+            _ => {}
+        }
+        self.assume_function_is_angelic = true;
```
wrwg: diag_level == RELAXED?

hermanventer: Probably, but I'll come back to this shortly.

PR #375, opened 2019-12-14
Specialize generic argument types

hermanventer:
## Description

Provide a visitor to specialize rustc types using the now available generic parameter to generic argument map from the caller.

Start by specializing actual argument types and fix up foreign contracts for VecDeque which broke a while back but went unnoticed because the break caused angelic behavior and there were no negative tests.

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/visitors.rs
```diff
@@ -3553,6 +3560,183 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
+    ) -> GenericArg<'tcx> {
+        match gen_arg.unpack() {
+            GenericArgKind::Type(ty) => {
+                self.specialize_generic_argument_type(ty, map, index).into()
+            }
+            _ => gen_arg,
+        }
+    }
+
+    fn specialize_substs(
```
wrwg: Ouch that we need to do this instead of getting it from MIR.

PR #376, opened 2019-12-16
Specialize non self generic arguments with caller map

hermanventer:
## Description

Extend specialization to all generic arguments of a callee. This enables more code to be resolved and analyzed, which fixes one test case but turn breaks two test cases. Since the latter probably passed only because of angelic behavior (and now fail because resolve panics), this still looks like progress in a forward direction.

## Type of change

## How Has This Been Tested?
./validate.sh
Running MIRAI on itself fails because of a resolve related panic.


checker/src/visitors.rs
```diff
@@ -531,35 +531,34 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
...
-                        if self.tcx.is_mir_available(resolved_def_id) {
-                            let mut devirtualized_call_info = call_info.clone();
-                            devirtualized_call_info.callee_def_id = resolved_def_id;
-                            return Some(
-                                self.create_and_cache_function_summary(&devirtualized_call_info),
-                            );
-                        }
+                // The environment of the caller provides a resolution context for the callee.
+                debug!("devirt gen_args {:?}", gen_args);
+                let param_env = self.tcx.param_env(self.def_id);
```
wrwg: Great

checker/tests/run-pass/for_in.rs
```diff
@@ -6,13 +6,13 @@
...
 
-#[macro_use]
-extern crate mirai_annotations;
-
-pub fn foo(n: usize) {
-    for ordinal in 2..=n {
-        verify!(ordinal - 1 >= 1);
-    }
-}
+//#[macro_use]
```
wrwg: Are those commented out tests stay like this? If so can you comment why and add a todo to reactivate.

hermanventer: I am actively working on the test failure, but it may take a number of PRs before I get there. A todo would not add value in this case.

PR #378, opened 2019-12-17
Introduce DiagLevel::PARANOID and run tests in this mode by default.

hermanventer:
## Description

DiagLevel::PARANOID will issue a diagnostic if an attempt is made to call a function that has no summary. The integration tests now run with this mode as the default.

In some test cases, this requires using test mode and annotating particular methods as the only methods to analyze, so that devirtualization can succeed.

## Type of change

## How Has This Been Tested?
./validate.sh

### DIFF
--- a/checker/src/options.rs
+++ b/checker/src/options.rs
@@ -5,6 +5,7 @@
 
 use clap::{App, AppSettings, Arg};
 use itertools::Itertools;
+use mirai_annotations::*;
 use rustc::session::config::ErrorOutputType;
 use rustc::session::early_error;
 use shellwords;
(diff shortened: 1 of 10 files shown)


checker/src/options.rs
```diff
@@ -28,7 +29,7 @@ fn make_options_parser<'a>() -> App<'a, 'a> {
         .long_help("Only #[test] methods and their usage are analyzed. This must be used together with the rustc --test option."))
     .arg(Arg::with_name("diag")
         .long("diag")
-        .possible_values(&["relaxed", "strict"])
+        .possible_values(&["relaxed", "strict", "paranoid"])
         .default_value("relaxed")
         .help("Level of diagnostics which will be produced.")
         .long_help("With `relaxed`, false positives will be avoided where possible. With `strict`, all errors will be reported."))
```
wrwg: Explain paranoid here

checker/tests/run-pass/contract_annotations.rs
```diff
@@ -6,6 +6,8 @@
 
 // Tests for annotations from the contracts crate.
 
+// MIRAI_FLAGS --test_only
```
wrwg: Don't see any test which uses --diag=paranoid?

checker/src/options.rs
hermanventer: I'll do this in a follow up PR, shortly.

PR #380, opened 2019-12-18
Fix generic specialization of <T, X<T>> style parameters

hermanventer:
## Description

Generic specialization did not properly specialize generic arguments that refer to earlier generic arguments, which led to the rust compiler's resolve method to panic.

This now seems fixed, but it just uncovers other problems, as usual. I'm putting this up as a separate PR to make reviewing easier.

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/visitors.rs
```diff
@@ -382,6 +382,9 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
     /// that is already in the cache.
     #[logfn_inputs(TRACE)]
     pub fn visit_body(&mut self, actual_args: &[(Rc<Path>, Rc<AbstractValue>)]) {
+        //        let mut stdout = std::io::stdout();
```
wrwg: Is this to stay? If so add a comment you keep for future debugging.

PR #381, opened 2019-12-19
Clean up command line option processing a bit

hermanventer:
## Description

Updated the text of the options to reflect the addition of "paranoid" to the --diag option. Also tweaked the formatting and hooked things up so that options supplied via the command line are recognized (and supersede options supplied via the MIRAI_FLAGS environment option.

Some care needs to be taken to not break scenarios where MIRAI is used a drop in replacement for rustc, which implies that command line options may be entirely for rustc and that "--" cannot be relied upon to separate them.

## Type of change

## How Has This Been Tested?
Ran mirai from the command line with various combinations of valid and invalid options.

### DIFF
--- a/checker/src/main.rs
+++ b/checker/src/main.rs
@@ -21,10 +21,11 @@ extern crate rustc;
 extern crate rustc_driver;
 extern crate rustc_interface;
 
-use log::info;
+use log::*;
 use mirai::callbacks;
 use mirai::options::Options;
 use mirai::utils;
+use mirai_annotations::*;
 use rustc::session::config::ErrorOutputType;
 use rustc::session::early_error;
 use std::env;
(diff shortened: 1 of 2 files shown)


checker/src/main.rs
```diff
@@ -42,69 +43,74 @@ fn main() {
         env_logger::init_from_env(e);
     }
 
-    // Get MIRAI options from environment variable. For now, we do not pass them
-    // on the command line because `RUSTC_WRAPPER=mirai cargo build` does not accept any
-    // options and arguments which it does not known about (even though we could remove them in
-    // this code here and not pass on to rustc). Once we have cargo integration
-    // of MIRAI we may want to change this.
+    // Get any options specified via the MIRAI_FLAGS environment variable
```
wrwg: Have you tried whether this actually worked if MIRAI is called via  RUSTC_WRAPPER=mirai cargo build as described in the comment you deleted? For me this actually failed because cargo doesn't seem to allow passing any arguments it doesn't know about. That's why I did not follow this pass in the first place. But experiment I made may have been faulty, so...

hermanventer: You are correct that cargo does not pass through command line arguments to rustc when building. For that something like MIRAI_FLAGS is needed and I have not changed anything there. This PR just adds the possibility of using mirai directly and passing in command line arguments as one would expect.

wrwg: Great, just wanted to be sure that it's clear this doesn't work. 

If we look at pure MIRAI command line (which we may want to use for creating `cargo mirai`), I'm not sure whether we need any more than `<mirai_args> -- <rustc_args>`, but good to have it now anyway.

PR #382, opened 2019-12-20
Promote preconditions for non root functions.

hermanventer:
## Description

The point of --test mode is to reduce false positives in the absence of explicit annotations. Consequently it is not desirable to complain that a public function does not have an explicit precondition annotation when the analyzer infers that such a condition is needed.

Instead, the precondition promotion now proceeds silently until the caller that fails to satisfy a precondition is an analysis root. In the case of --test, this will be functions annotated with #[test]. Without this flag, it will be functions that are public.

## Type of change

## How Has This Been Tested?
./validate.sh


checker/tests/run-pass/test_source.rs
```diff
@@ -31,10 +31,9 @@ fn no_summary_analyzed_anyway() {
         }
     }
     let d: &dyn Dynamic = &S {} as &dyn Dynamic; // forget type info of S
-    verify!(d.f(1) == 3); // normally, this statement would disable verification for this
-                          // function and we would not see the message below. With --diag=strict,
-                          // we do not see an error here (as d.f is uninterpreted), but we still
-                          // see the below error.
+
+    // With --diag=strict we still see an error even though d.f is uninterpreted.
```
wrwg: That kind of reverses the purpose of this test. It's purpose was to test that there is no diag for the uninterpreted function call, but still for others in this function.

Issue #390, opened 2020-01-12
Panic when MIRAI checks itself: rustc `DefId` as `Hasher` unimplemented


phlip9:
## Issue

Hey! I was playing around with MIRAI this weekend, but I hit an issue trying to get MIRAI to check itself on the latest master. After poking around a bit, it looks like there's a foreign contract or blacklist missing? Specifically, it looks like MIRAI is complaining about something `Unimplemented` for `rustc::hir::def_id::DefId as std::hash::Hasher`. I'm sorry if that's a bit vague; I'm not very familiar with MIRAI :)

## Steps to Reproduce

Run `validate.sh` on current master (c4c58ad6e79779ef00d146aac72d9f313908f496). I also tested on the previous two commits; same problem.

## Results

For completeness, here's the part of log when run with `MIRAI_LOG=trace`: https://gist.github.com/phlip9/6810fe0d4bc829c7de69a36df798628f

Otherwise, here's the relevant part of `validate.sh` where MIRAI checks itself:

```sh
~/dev/MIRAI$ RUSTFLAGS="-Z always_encode_mir" RUSTC_WRAPPER=mirai RUST_BACKTRACE=1 MIRAI_LOG=info cargo build --lib -p mirai
   Compiling mirai v1.0.5 (/Users/philiphayes/dev/MIRAI/checker)
[2020-01-12T01:57:08Z INFO  mirai] MIRAI options from environment: Options { single_func: None, test_only: false, diag_level: RELAXED }
[2020-01-12T01:57:08Z INFO  mirai] MIRAI options modified by command line: Options { single_func: None, test_only: false, diag_level: RELAXED }
[2020-01-12T01:57:08Z INFO  mirai::callbacks] Processing input file: checker/src/lib.rs
[2020-01-12T01:57:13Z INFO  mirai::callbacks] storing summaries for checker/src/lib.rs at /var/folders/y7/nfv_8ryx2x78f1gvdv_8rfd1rynvd9/T/mirai_temp_dir.CiiLgkkjcSR1/.summary_store.sled
[2020-01-12T01:57:13Z INFO  mirai::summaries] creating a non default new summary store
[2020-01-12T01:57:13Z INFO  mirai::callbacks] skipping function mirai.abstract_value._IMPL_SERIALIZE_FOR_AbstractValue as it is not public
# ... snip ...
[...]

PR #395, opened 2020-01-16
Handle static variables that are structured values

hermanventer:
## Description

Static variables can be structured values and the functions that compute them can thus have summaries that side effect (construct) the result variable. This was not handled in the code that initializes static variables from summaries.

Related to #390 

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/visitors.rs
```diff
@@ -462,6 +462,8 @@ impl<'analysis, 'compilation, 'tcx, E> MirVisitor<'analysis, 'compilation, 'tcx,
                 let param_path = Path::new_local(i + 1);
                 environment.value_map = environment.value_map.insert(param_path, arg.clone());
             }
+            //todo: if the argument is structured and one of the elements is a function constant,
+            //add that to the environment as well.
```
wrwg: Space after //

PR #403, opened 2020-01-24
Check in summary_key_str that “num” belongs to crate “core”.

vakaras:
## Description

The [summary_key_str](https://github.com/facebookexperimental/MIRAI/blob/master/checker/src/utils.rs#L303) function incorrectly assumes that if a module is named `num`, it comes from the `core` crate. This assumption is, for example, violated by the [lexical-core](https://github.com/Alexhuszagh/rust-lexical/blob/master/lexical-core/src/util/num.rs) crate, which causes the call to `tcx.type_of(parent_def_id)` to panic. This pull request adds a check that the actual crate to which `num` belongs is `core`.

## Type of change

## Checklist:

- [ ] Ensure the test suite passes.
- [ ] If you haven't already, complete your CLA here: <https://code.facebook.com/cla>


hermanventer:
Thanks for fixing this.

PR #406, opened 2020-01-28
Provide way to specify --single_func via an environment parameter

hermanventer:
## Description

Provide way to specify --single_func via an environment parameter. This restores a small convenience that I use in my own work flow.

Also update the README to know about the 'paranoid' strictness level.

Finally tweak utils::summary_key_str to not produce name strings with ".." in them because those cannot be overridden by foreign contracts.

## Type of change

## How Has This Been Tested?
./validate.sh


checker/src/options.rs
```diff
@@ -86,6 +86,8 @@ impl Options {
     /// Parses options from a list of strings. Any content beyond the leftmost `--` token
     /// will be returned (excluding this token).
     pub fn parse(&mut self, args: &[String]) -> Vec<String> {
+        self.single_func = std::env::var("MIRAI_SINGLE_FUNC").ok();
```
wrwg: Can't use MIRAI_FLAGS=--single_func=name or similar? I think this was the substitute when it was removed (based on your feedback to not add a ton of env vars.)

hermanventer: Ah, good point. I've been a bit too set in my ways.

PR #424, opened 2020-02-07
Extract thin pointer from first field of fat pointer via union

hermanventer:
## Description

Support accessing the pointer part of a fat pointer via a union where the other case is just a pointer. To do this, it is not enough to just use a common field number for both cases, since the fat pointer will store its pointer in a qualified path.

Also updated the black list by removing crates that work now and annotating the remainder with the reasons they are black listed.

## Type of change

## How Has This Been Tested?
./validate.sh with added test case
ran MIRAI over Libra


checker/src/callbacks.rs
```diff
@@ -161,47 +161,28 @@ impl MiraiCallbacks {
...
-            || file_name.contains("common/executable-helpers/src")
-            || file_name.contains("client/cli/src")
-            || file_name.contains("consensus/safety-rules/src")
-            || file_name.contains("language/compiler/bytecode-source-map/src")
-            || file_name.contains("language/compiler/ir-to-bytecode/syntax/src")
-            || file_name.contains("common/metrics/src")
-            || file_name.contains("config/config-builder/src")
-            || file_name.contains("secure/storage/vault/src")
-            || file_name.contains("secure/net/src")
+        file_name.contains("secure/net/src") // false positives
```
wrwg: Wow that is great to see the blacklist shrink by more than 50%.

PR #426, opened 2020-02-08
Foreign contracts for arith operations on usize operands

hermanventer:
## Description

Some of these are needed to make progress with supporting vector operations, so I just did all of those that involve usize operands. This can serve as a template for other integer types.

Testing all this for all corner cases is more than I can do now, so a careful reading will be much appreciated.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


standard_contracts/src/foreign_contracts.rs
```diff
@@ -938,57 +938,169 @@ pub mod core {
...
+            }
+
             pub fn sub_with_overflow<T>(x: T, y: T) -> (T, bool) {
                 result!()
             }
+            pub fn sub_with_overflow__usize(x: usize, y: usize) -> (usize, bool) {
+                let result = (x as i128) + (-(y as i128));
+                (
+                    (result % ((std::usize::MAX as i128) + 1)) as usize,
+                    result > (std::usize::MAX as i128),
```
wrwg: This looks somehow wrong. Shouldn't subtraction overflow when result < 0?

standard_contracts/src/foreign_contracts.rs
```diff
@@ -938,57 +938,169 @@ pub mod core {
...
+                )
+            }
+
+            /// Performs an exact division, resulting in undefined behavior where
```
wrwg: s/where/when

PR #471, opened 2020-03-27
Simplify some common patterns

hermanventer:
## Description

As the analysis gets more precise and carries on for longer before giving up, expressions can grow beyond the k-limit of 1000 terms. At this point such expressions become TOP and the analysis loses precision and spews false positives. Such expressions also make debugging difficult.

Since a lot of the terms in such expressions are highly stylized they can be efficiently simplified with heuristic pattern matching rules. As new rules become useful, I add them the existing set, hence this PR.

Of note when reviewing these rules: The expressions are functional and are not expected to fail. I.e. all side effects and range checks have already been performed elsewhere.

The rules are somewhat subtle and easy to get wrong, so careful review will be much appreciated.

Also included here is a "fix" that stops the analyzer from crashing in a situation that ought not arise.

## Type of change

## How Has This Been Tested?
./validate.sh
cargo test

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -307,6 +307,9 @@ impl AbstractValueTrait for Rc<AbstractValue> {
                 }
             }
         }
+        if let Expression::Neg { operand } = &other.expression {
+            return self.subtract(operand.clone());
+        }
         self.try_to_simplify_binary_op(other, ConstantDomain::add, Self::addition, |l, r| {
             AbstractValue::make_binary(l, r, |left, right| Expression::Add { left, right })
         })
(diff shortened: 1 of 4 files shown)


checker/src/abstract_value.rs
```diff
@@ -574,6 +577,35 @@ impl AbstractValueTrait for Rc<AbstractValue> {
                 .cast(target_type.clone())
                 .join(right.cast(target_type), &path),
             _ => {
+                match &self.expression {
+                    // [(x as t1) as target_type] -> x as target_type if t1.max_value() >= target_type.max_value()
```
wrwg: Perhaps add `where t1 = any int` or something?

hermanventer:
MIRAI makes very sparing use of Z3 and none at all during expression refinement. This is a design choice that has to be made long before it can be informed by data. In experience thus far, heuristics work well enough for real life code. I hope this will remain the case, but we'll have to see.

PR #525, opened 2020-05-28
Clarify how to run a locally built mirai.

stonebuddha:
## Description

The "Running" section of the developer guide has mixed instructions for using both a locally built binary and a globally installed binary, on both a single file (as rustc) and a crate (via cargo).
It also seems unnecessary to have instructions for globally installed binary, because they are covered somewhere else, and a developer just needs the locally built one during development.

This commit separates the instructions for single files from those for crates, updates all the commands to use a locally built binary, and tweaks the indentation in debug configurations.

## Type of change

None of the above: only documentation change.

## How Has This Been Tested?

Followed the instructions to build and run locally built mirai on both single files and crates.

### DIFF
--- a/documentation/DeveloperGuide.md
+++ b/documentation/DeveloperGuide.md
@@ -39,7 +39,9 @@ Alternatively there is also
 ## Running
 
 Running `cargo build` will produce a binary at `target/debug/mirai`.
- 
+
+### On a single file
+
 Unfortunately cargo build sets the dynamic library load path (rpath) that is linked into the binary to a path that is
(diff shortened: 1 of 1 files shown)


hermanventer:
I presume that you have actually carried out the process documented by this. If so, that is your test plan. If not, please do so.

stonebuddha:
> I presume that you have actually carried out the process documented by this. If so, that is your test plan. If not, please do so.

Yeah, I've followed the process on my machine. I'll update the test plan.

PR #527, opened 2020-06-01
Simplify *& in Path factory, make &str thin

hermanventer:
## Description

Make path refinement less hacky by moving most *& elimination into the Path::new_qualified factory function.

Also model &str pointers, correctly, as thin pointers rather than as slice pointers. Deal with the fall out from this change.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -312,7 +312,6 @@ pub trait AbstractValueTrait: Sized {
     fn is_bottom(&self) -> bool;
     fn is_compile_time_constant(&self) -> bool;
     fn is_contained_in_zeroed_heap_block(&self) -> bool;
-    fn is_path_alias(&self) -> bool;
     fn is_top(&self) -> bool;
     fn join(&self, other: Self, path: &Rc<Path>) -> Self;
     fn less_or_equal(&self, other: Self) -> Self;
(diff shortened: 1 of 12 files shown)


checker/src/abstract_value.rs
```diff
@@ -2017,6 +1997,31 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+            Expression::Reference(..) => true,
+            Expression::UninterpretedCall { path, .. }
+            | Expression::Variable { path, .. }
+            | Expression::Widen { path, .. } => {
+                if let PathEnum::Alias { value } = &path.value {
+                    return value.refers_to_unknown_location();
+                }
+                false
+            }
+            _ => false,
```
stonebuddha, line 2021: Does this function also need to consider `Join` and `Switch` expressions, since they are similar to `ConditionalExpression`?

hermanventer: Excellent question. I think the answer is yes.

stonebuddha:
I've seen
```
analysis of parking_lot.raw_rwlock.implement_raw_rwlock.lock_shared_slow.closure timed out after 3 seconds
```
when running mirai on itself. Is the timeout acceptable?

hermanventer:
Not really, but this is not currently a check-in gate. Errors like these are often latent and masked by other errors. Fixing everything at once is not feasible or desirable.

PR #532, opened 2020-06-05
Introduce a tag domain and keep track of it in abstract values

stonebuddha:
## Description

This commit is a follow-up of #513: provide annotations to add and check type tags on arbitrary values, and implement a tag domain whose elements are maps from type tags to lifted Booleans. A tag mapped to `True` must be present, a tag mapped to `False` must be absent, and a tag mapped to `Top` can be either present or absent. Tags would not be removed and tagged values would not be mutable.

To control the behavior of transfer functions for different tags, we restrict tag types to generic types `Tag<SET, ...>` whose first parameter should be a constant of type `mirai_annotations::TagPropagationSet`, which is used as a bit-vector to control whether the tag would be propagated by an instruction or not. The type `mirai_annotations::TagPropagationSet` is an alias of `u128`. Propagation sets can be created via the `mirai_annotations::tag_propagation_set!` macro, which takes a list of values of the enum type `mirai_annotations::TagPropagation`.

Transfer functions for tags have not been implemented. Nevertheless, this commit includes a preliminary implementation for adding/checking tags, and adds a new test for an "add-and-then-check" scenario.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra


annotations/src/lib.rs
```diff
@@ -44,7 +69,22 @@ macro_rules! add_tag {
...
         if cfg!(mirai) {
-            mirai_annotations::mirai_has_tag::<$tag>($value)
+            mirai_annotations::mirai_has_tag::<_, $tag>($value)
+        } else {
+            true
+        }
+    };
+}
+
+/// Provides a way to check if a value has *never* been tagged with a type using add_tag!.
```
hermanventer, line 79: *never* -> not

annotations/src/lib.rs
```diff
@@ -44,7 +69,22 @@ macro_rules! add_tag {
...
+/// When compiled with an unmodified Rust compiler, this results in true.
+/// When compiled with MIRAI, this will be true if all data flows into the argument of this
```
hermanventer, line 81: all -> none

annotations/src/lib.rs
```diff
@@ -44,7 +69,22 @@ macro_rules! add_tag {
...
+/// call has *never* gone via a call to add_tag!.
```
hermanventer, line 82: *never* ->

annotations/src/lib.rs
```diff
@@ -22,6 +25,28 @@ macro_rules! abstract_value {
...
+pub trait MiraiTagKind {}
+
+/// A generic tag type to allow user-customizable tag kind and control mask.
+pub struct MiraiTag<KIND: MiraiTagKind, const MASK: u128> {
+    kind: std::marker::PhantomData<*const KIND>,
+}
+
+/// A trait to identify MIRAI tag types. The trait is sealed and is not meant to be implemented
+/// outside of the crate `mirai_annotations`.
+pub trait MiraiSealedTag: private::Sealed {}
```
hermanventer, line 38: Instead of using the type system, in this rather convoluted way, to enforce the requirement that every tag type specifies its transfer behavior, we could do the checking in MIRAI. I.e. keep the existing design, which is that a tag can be any kind of type, but give a MIRAI diagnostic if the type does not have a generic parameter that is a constant  of type TransferMask defined as an alias for u128 in MIRAI Annotations.

annotations/src/lib.rs
```diff
@@ -961,11 +1001,17 @@ pub fn mirai_abstract_value<T>(x: T) -> T {
...
 // Helper function for MIRAI. Should only be called via the has_tag! macro.
 #[doc(hidden)]
-pub fn mirai_has_tag<V, T>(_v: V) -> bool {
+pub fn mirai_has_tag<V, T: MiraiSealedTag>(_v: &V) -> bool {
+    false
+}
+
+// Helper function for MIRAI. Should only be called via the not_has_tag! macro.
+#[doc(hidden)]
+pub fn mirai_not_has_tag<V, T: MiraiSealedTag>(_v: &V) -> bool {
```
hermanventer, line 1014: Rename to mirai_does_not_have_tag

checker/src/block_visitor.rs
```diff
@@ -1749,6 +1749,10 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
                                 if let Some(arg_val) = gen_args.as_ref().get(*index as usize) {
                                     return self.visit_constant(None, arg_val.expect_const());
                                 }
+                            } else {
```
hermanventer, line 1752: This can go away if you get rid of Sealed. There is still a problem here, but we need a different test case for it (and and issue, at some point).

checker/src/call_visitor.rs
```diff
@@ -1948,4 +2099,39 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+                        {
+                            let tag_kind = adt_def.did;
+                            let tag_mask = *data;
+                            return (tag_kind, tag_mask);
+                        }
+                    }
+                }
+            }
+        }
+        assume_unreachable!();
```
hermanventer, line 2135: Always justify assumptions via a comment. And NEVER assume that user input is correct.

checker/src/abstract_value.rs
```diff
@@ -45,7 +46,10 @@ pub struct AbstractValue {
     pub expression_size: u64,
     /// Cached interval domain element computed on demand by get_as_interval.
     #[serde(skip)]
-    interval: RefCell<Option<Rc<IntervalDomain>>>,
+    pub interval: RefCell<Option<Rc<IntervalDomain>>>,
+    /// Cached tag domain element computed on demand by get_tags.
+    #[serde(skip)]
+    pub tags: RefCell<Option<Rc<TagDomain>>>,
```
hermanventer, line 52: I think you want tags to be a map from tag type to tag value, where the latter is a lifted Boolean. (I.e. a domain whose elements are {}, {false}, {true}, {false, true} with BOTTOM={} (as always) and TOP={false, true}).

checker/src/call_visitor.rs
```diff
@@ -1994,4 +2127,51 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
         }
         Rc::new("dummy argument".to_string())
     }
+
+    /// Extract the tag kind and the control mask from the generic arg of the function call
+    /// underlying `add_tag!` or `has_tag!`. The tag type should be the second generic argument
+    /// of the current function call. The tag type itself should also be parametrized, and its
+    /// first type parameter should be a constant of type `u128`, which represents the control
+    /// mask. Return a pair of the name of the tag type, as well as the control mask.
```
hermanventer, line 2135: `of type EnumSet<TagPropagation>`

checker/src/call_visitor.rs
```diff
@@ -1994,4 +2127,51 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+            self.callee_known_name == KnownNames::MiraiAddTag
+                || self.callee_known_name == KnownNames::MiraiHasTag
+                || self.callee_known_name == KnownNames::MiraiDoesNotHaveTag
+        );
+        if let Some(rustc_gen_args) = self.callee_generic_arguments {
+            checked_assume!(rustc_gen_args.len() == 2);
+            if let GenericArgKind::Type(tag_rustc_type) = rustc_gen_args[1].unpack() {
+                if let TyKind::Adt(adt_def, substs_ref) = &tag_rustc_type.kind {
+                    checked_assume!(substs_ref.len() > 0);
+                    if let GenericArgKind::Const(tag_mask_rustc_const) = substs_ref[0].unpack() {
```
hermanventer, line 2147: You can check the ty field of tag_mask_rustc_const to ensure that the constant is actually of type `EnumSet<TagPropagation>`. If not, you can generate a diagnostic here.

stonebuddha:
I have to increase `MAX_ANALYSIS_TIME_FOR_BODY` from 3 to 5 to pass the check. The travis build times out on the test `bit_vector_and_vec_push`.

hermanventer:
That is OK for the sake of moving forward, but it is worrying that something in this PR slows things down. We probably don't want that, but understanding the issue may take some time. It could be that addition of another field to AbstractValue is the root cause.

PR #536, opened 2020-06-12
Remove deref selector when using Path::implicit_dereference

stonebuddha:
## Description

The function `Path::implicit_deference` is used when splitting path qualifiers while doing weak updates. If the qualifier of a qualified path is indeed canonicalized through this function, the selector should also be removed when it is `PathSelector::Deref`. This commit adds a test to check if implicit dereference works properly.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/path.rs
+++ b/checker/src/path.rs
@@ -92,23 +92,31 @@ impl Path {
     /// they are already canonical and because doing so can lead to recursive loops as refined paths
     /// re-introduce joined paths that have been split earlier.
     /// A split path, however, can be serve as the qualifier of a newly constructed qualified path
-    /// and this path might not be canonical. This routine removes the two sources of
+    /// and this path might not be canonical. This routine tries to remove the two sources of
     /// de-canonicalization that are currently know. Essentially: when a path that binds to a value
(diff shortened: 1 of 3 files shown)


checker/src/path.rs
```diff
@@ -98,17 +98,17 @@ impl Path {
     /// be the one without the reference, or the the actual heap block, if the path binds to a heap
     /// location.
     #[logfn_inputs(DEBUG)]
-    pub fn implicit_dereference(path: Rc<Path>, environment: &Environment) -> Rc<Path> {
+    pub fn implicit_dereference(path: Rc<Path>, environment: &Environment) -> Option<Rc<Path>> {
```
hermanventer, line 101: This can probably be reference (&Rc<Path>). Also the name of this function (and its comments) now seem a bit misleading.

stonebuddha:
> Nice catch. Before merging can you try to make things a bit more readable?

I added a few comments. Do they clarify things well?

hermanventer:
Only slightly. I'll let you merge this but please remind me to try and fix the comments.

PR #538, opened 2020-06-15
Use param_env with reveal_all so that trait methods resolve better

hermanventer:
## Description

When resolving trait methods, the param env should be flagged with reveal_all for some kinds of resolve calls. Fixing this uncovered problems with the type specialization of Opaque types and Projection types (also very Trait specific stuff). Also clarified some closure specific code.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI on Libra

### DIFF
--- a/checker/src/type_visitor.rs
+++ b/checker/src/type_visitor.rs
@@ -64,7 +64,7 @@ impl<'analysis, 'compilation, 'tcx> TypeVisitor<'tcx> {
     /// additional parameters. We pre-populate the environment with entries for these because
     /// there is no convenient way to look up their types later on. I.e. unlike ordinary parameters
     /// whose types can be looked up in mir.local_decls, these extra parameters need their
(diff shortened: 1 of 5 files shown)


checker/src/type_visitor.rs
```diff
@@ -60,8 +60,13 @@ impl<'analysis, 'compilation, 'tcx> TypeVisitor<'tcx> {
         self.path_ty_cache = HashMap::default();
     }
 
+    /// Parameters that are closures bring enclosed variables with them that are effectively
+    /// additional parameters. We pre-populate the environment with entries for these because
+    /// there is no convenient way to look up their types later on. I.e. unlike ordinary parameters
+    /// whose types can be looked up in mir.local_decls, these extra parameters need their
+    /// types extracted from the from the closure type definitions via the tricky logic below.
```
stonebuddha, line 67: from the from the -> from the

PR #540, opened 2020-06-16
move_or_copy_elements should do weak updates in all cases

### LINKED ISSUE #537: False bug reported when moving things to a target with multiple possible paths
## Issue

MIRAI reports a false bug when moving things to a target with multiple possible paths. The cause might be that the function `BodyVisitor::copy_or_move_elements` should expect unique target paths.

## Steps to Reproduce

Run MIRAI on the following code:
``` rust
#[macro_use]

struct Bool {
    content: bool,
}

pub fn test(cond: bool) {
    let mut left = Bool { content: false };
    let mut right = Bool { content: false };
    let join;
    if cond {
        join = &mut left;
    } else {
        join = &mut right;
    }
    *join = Bool { content: true };
    verify!(left.content || right.content);
}

pub fn main() {}
```

## Expected Behavior

The program contains no bugs, so MIRAI is supposed to report no errors.

## Actual Results

MIRAI reports a provably false verification condition:
```
error: provably false verification condition
  --> src/main.rs:18:5
   |
18 |     verify!(left.content || right.content);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

## Environment

rustc 1.45.0-nightly (5fd2f06e9 2020-05-31)

hermanventer:
## Description

Use update_value_at more consistently so that weak updates happen.

Fixes #537

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/body_visitor.rs
+++ b/checker/src/body_visitor.rs
@@ -1491,16 +1491,14 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
         }
 
         // Get here for paths that are not patterns.
-        let mut value_map = self.current_environment.value_map.clone();
+        let value_map = self.current_environment.value_map.clone();
         let is_closure = matches!(&root_rustc_type.kind, TyKind::Closure(..));
         let value = self.lookup_path_and_refine_result(source_path.clone(), root_rustc_type);
         let val_type = value.expression.infer_type();
         let mut no_children = true;
         if val_type == ExpressionType::NonPrimitive || is_closure {
             // First look at paths that are rooted in rpath.
-            for (path, value) in self
-                .current_environment
-                .value_map
+            for (path, value) in value_map
                 .iter()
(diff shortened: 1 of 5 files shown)


checker/src/body_visitor.rs
```diff
@@ -1498,21 +1498,21 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
...
+                        self.current_environment.value_map.remove(path);
                 } else {
                     trace!("copying child {:?} to {:?}", value, qualified_path);
                 };
-                value_map = value_map.insert(qualified_path.clone(), value.clone());
+                self.current_environment
+                    .update_value_at(qualified_path, value.clone());
                 no_children = false;
             }
         }
```
stonebuddha, line 1518: On line 1524 & line 1537, should we operate directly on `self.current_environment.value_map` instead of on `value_map`?

hermanventer: Indeed. Good catch.

PR #541, opened 2020-06-17
Implement transfer functions for the tag domain

stonebuddha:
## Description

This commit implements transfer functions for the tag domain (introduced in #532). Similar to the interval domain, tags are computed on demand via function `AbstractValue::get_tags`. The transfer function looks up each tag's propagation set (of type `mirai_annotations::TagPropagationSet`) to decide whether an expression should propagate the tag from sub-expressions or not.

The function `BodyVisitor::attach_tag_to_elements` and `Environment::update_value_with_tag` implement the tag adding routine. They are similar to `BodyVisitor::copy_or_move_elements` and `Environment::update_value_at`, but they don't enumerate possible paths rooted at the tagged value. On the other hand, when checking if a value--located at a qualified path--has a tag, one has to also check if its qualifier has been attached with a tag. The mechanism is implemented by function `CallVisitor::handle_has_tag`.

Caveats: weak updates for tags are not implemented; expression simplifications (which happen in methods of `AbstractValueTrait`) may lose track of tags.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -307,6 +307,26 @@ impl AbstractValue {
     pub fn make_typed_unknown(var_type: ExpressionType, path: Rc<Path>) -> Rc<AbstractValue> {
         AbstractValue::make_from(Expression::Variable { path, var_type }, 1)
     }
+
+    /// Creates an abstract value that is basically the same as the given value, except that
+    /// a tag domain element is attached to it. This routine is useful when a value is being
+    /// refined and we don't want to drop the existing tags associated with that value.
+    /// todo: this will change to a new tagged expression.
+    #[logfn_inputs(TRACE)]
+    pub fn make_with_tags(
+        value: &Rc<AbstractValue>,
+        tags_opt: Option<&Rc<TagDomain>>,
+    ) -> Rc<AbstractValue> {
+        match tags_opt {
+            None => value.clone(),
+            Some(tags) => Rc::new(AbstractValue {
+                expression: value.expression.clone(),
+                expression_size: value.expression_size,
+                interval: RefCell::new(value.interval.borrow().clone()),
+                tags: RefCell::new(Some(tags.clone())),
+            }),
+        }
+    }
 }
 
 pub trait AbstractValueTrait: Sized {
@@ -332,6 +352,7 @@ pub trait AbstractValueTrait: Sized {
     fn equals(&self, other: Self) -> Self;
     fn greater_or_equal(&self, other: Self) -> Self;
     fn greater_than(&self, other: Self) -> Self;
+    fn has_tag(&self, tag: &Tag) -> Rc<AbstractValue>;
     fn implies(&self, other: &Self) -> bool;
     fn implies_not(&self, other: &Self) -> bool;
     fn intrinsic_binary(&self, other: Self, name: KnownNames) -> Self;
@@ -502,7 +523,7 @@ impl AbstractValueTrait for Rc<AbstractValue> {
     /// Returns an element which marks `tag` whose presence is indicated by `val`.
     #[logfn_inputs(TRACE)]
     fn add_tag(&self, tag: Tag, val: BoolDomain) -> Rc<AbstractValue> {
-        let old_tags = self.get_tags();
+        let old_tags = self.get_cached_tags();
         let new_tags = old_tags.add_tag(tag, val);
         Rc::new(AbstractValue {
             expression: self.expression.clone(),
@@ -1082,7 +1103,7 @@ impl AbstractValueTrait for Rc<AbstractValue> {
     // generic types when constructed, but then leak out to caller contexts via summaries.
     #[logfn_inputs(TRACE)]
     fn try_to_retype_as(&self, target_type: &ExpressionType) -> Rc<AbstractValue> {
-        match &self.expression {
+        let result = match &self.expression {
             Expression::Add { left, right } => left
                 .try_to_retype_as(target_type)
                 .addition(right.try_to_retype_as(target_type)),
@@ -1148,7 +1169,9 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             Expression::Widen { .. } => self.clone(),
 
             _ => self.clone(),
-        }
+        };
+        // todo: unnecessary with tagged expressions
+        AbstractValue::make_with_tags(&result, self.tags.borrow().as_ref())
     }
 
     /// Returns an element that is "*self".
@@ -1405,6 +1428,12 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             left,
             right,
         })
+    }
+
+    /// Returns an abstract value that describes whether `tag` is attached to `self` or not.
+    #[logfn_inputs(TRACE)]
+    fn has_tag(&self, tag: &Tag) -> Rc<AbstractValue> {
+        Rc::new(self.get_cached_tags().has_tag(tag).into())
     }
 
     /// Returns true if "self => other" is known at compile time to be true.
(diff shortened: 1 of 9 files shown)


checker/src/abstract_value.rs
```diff
@@ -307,6 +295,25 @@ impl AbstractValue {
...
+    /// refined and we don't want to drop the existing tags associated with that value.
+    #[logfn_inputs(TRACE)]
+    pub fn make_with_tags(
+        value: &Rc<AbstractValue>,
+        tags_opt: Option<&Rc<TagDomain>>,
+    ) -> Rc<AbstractValue> {
+        match tags_opt {
+            None => value.clone(),
+            Some(tags) => Rc::new(AbstractValue {
+                expression: value.expression.clone(),
```
hermanventer, line 310: As discussed earlier, this will change to a new expression tag in a future PR.

checker/src/abstract_value.rs
```diff
@@ -332,6 +339,7 @@ pub trait AbstractValueTrait: Sized {
     fn equals(&self, other: Self) -> Self;
     fn greater_or_equal(&self, other: Self) -> Self;
     fn greater_than(&self, other: Self) -> Self;
+    fn has_tag(&self, tag: &Tag) -> Option<Rc<AbstractValue>>;
```
hermanventer, line 342: I find myself wondering why the return result is optional. It seems that a lifted Boolean can always be returned.

checker/src/abstract_value.rs
```diff
@@ -1148,7 +1156,8 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             Expression::Widen { .. } => self.clone(),
 
             _ => self.clone(),
-        }
+        };
+        AbstractValue::make_with_tags(&result, self.tags.borrow().as_ref())
```
hermanventer, line 1160: This will become unnecessary with a dedicated kind of expression for tagged values.

checker/src/abstract_value.rs
```diff
@@ -1407,6 +1416,19 @@ impl AbstractValueTrait for Rc<AbstractValue> {
         })
     }
 
+    /// Returns an abstract value that describes whether `tag` is attached to `self` or not.
+    /// If we cannot decide, return None.
+    #[logfn_inputs(TRACE)]
+    fn has_tag(&self, tag: &Tag) -> Option<Rc<AbstractValue>> {
+        let tags = self.get_cached_tags();
+        match tags.has_tag(tag) {
+            BoolDomain::Top => None,
```
hermanventer, line 1425: This may as well be TOP. See also AbstractValue::as_bool_if_known.

checker/src/abstract_value.rs
```diff
@@ -2537,16 +2559,126 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                return TagDomain::for_empty_set();
+            }
+
+            Expression::ConditionalExpression {
+                condition,
+                consequent,
+                alternate,
+            } => {
+                // tags(condition) | (tags(consequent) join tags(alternate))
+                return condition.get_cached_tags().union(
```
hermanventer, line 2583: I'm a bit surprised to see get_cached_tags here. I fully expected to see get_tags. Can you explain why?

checker/src/call_visitor.rs
```diff
@@ -1035,31 +1035,29 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
     #[logfn_inputs(TRACE)]
     fn handle_add_tag(&mut self) {
         precondition!(self.actual_args.len() == 1);
-        if let Some(tag) = self.extract_tag_kind_and_mask() {
+
+        if let Some(tag) = self.extract_tag_kind_and_propagation_set() {
```
hermanventer, line 1039: You can deconstruct tag here so that you don't have to do tag.0 and tag.1 below.

checker/src/call_visitor.rs
```diff
@@ -2157,15 +2179,15 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
-                    &tag_mask_abs.expression
+                    &tag_propagation_set_abs.expression
                 {
                     let tag_kind = tag_adt_def.did;
-                    let tag_mask = *data;
-                    Some((tag_kind, tag_mask))
+                    let tag_propagation_set = *data;
+                    Some((tag_kind, tag_propagation_set))
                 } else {
                     assume_unreachable!(
```
hermanventer, line 2192: Looking at this again, I am wondering if this is actually unreachable. Can you add a comment to convince that this can never be reached?

checker/src/call_visitor.rs
```diff
@@ -1190,49 +1188,72 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
-                .current_environment
-                .update_value_at(target_path, result);
+            self.block_visitor.bv.current_environment.update_value_at(
+                target_path.clone(),
+                result.unwrap_or_else(|| {
+                    AbstractValue::make_typed_unknown(ExpressionType::Bool, target_path)
+                }),
+            );
         } else {
             assume_unreachable!();
```
hermanventer, line 1258: Please motivate your assumption.

checker/src/call_visitor.rs
```diff
@@ -1190,49 +1188,72 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
-            let source_abs = self
-                .block_visitor
-                .bv
-                .lookup_path_and_refine_result(source_path, source_rustc_type);
-            let source_tags = source_abs.get_cached_tags();
+            // Lookup the value located at the source path and check its associated tags.
+            // Also lookup the ancestors of the source path to find possible tags.
+            let mut cur_path = source_path;
+            let mut cur_result = Rc::new(abstract_value::FALSE);
+            loop {
```
hermanventer, line 1208: In other parts of MIRAI the pattern is to loop through all of the (path, value) pairs in the current environment, filtering for paths that are rooted in source_path. You can return from the loop as soon as one of the qualifying (path, value) pairs has value with the tag. Use iterators.

checker/src/call_visitor.rs
```diff
@@ -1190,49 +1188,72 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
     #[logfn_inputs(TRACE)]
     fn handle_has_tag(&mut self, checking_presence: bool) {
         precondition!(self.actual_args.len() == 1);
-        let result: Rc<AbstractValue>;
-        if let Some(tag) = self.extract_tag_kind_and_mask() {
+
+        let result: Option<Rc<AbstractValue>>;
+        if let Some(tag) = self.extract_tag_kind_and_propagation_set() {
             let source_path = Path::new_deref(self.actual_args[0].0.clone())
```
hermanventer, line 1194: The deref and the refine seems unnecessary at best and possibly wrong.

checker/src/abstract_value.rs
stonebuddha: Added a todo comment.

checker/src/abstract_value.rs
stonebuddha: Added todo comments.

checker/src/abstract_value.rs
stonebuddha: There are two reasons:
1. The subexpressions might already have their tag field computed, so we don't need to recompute from scratch.
2. For now there is no tagged expressions, so it seems impossible to reconstruct tags from the expression itself.

checker/src/call_visitor.rs
stonebuddha: `tag.0` and `tag.1` are only used in logging, and `tag` as a whole is used later in the code.

checker/src/call_visitor.rs
stonebuddha: I also tried this approach: when adding tags, loop through all of the (path, value) pairs to update the values. An issue for this approach is that some of the paths may not be in the current environment, e.g., struct fields of a function parameter.

checker/src/call_visitor.rs
stonebuddha: The argument of `add_tag!` and `has_tag!` is a reference. What is a correct way to dereference a path?

checker/src/abstract_value.rs
hermanventer: 1. I confused get_cached_tags with get_tags, so all is good.

checker/src/call_visitor.rs
hermanventer: The logging is likely remain in the code, so it may as well be as readable as possible.

checker/src/call_visitor.rs
hermanventer: OK, I guess it is complicated, see below.

checker/src/call_visitor.rs
hermanventer: This is where things get complicated. Essentially, all decisions must be postponed until the right environment is present. Let's discuss this over VC.

checker/src/call_visitor.rs
stonebuddha: Changed the type of tags from pairs to a structs. Now the fields are named.

hermanventer:
Let's get this merged and resolve the outstanding issues in later PRs.

PR #542, opened 2020-06-18
Introduce a new expression kind to record tags explicitly

stonebuddha:
## Description

This commit adds a new variant `TaggedExpression` to the type `Expression`. Semantically, `TaggedExpression { operand, tag }` is almost the same as `operand`, except that `tag` is attached to `operand`. This new variant is used to explicitly keep track of tags, and also to follow a convention for `AbstractValue`: when two values have the same expression, they can be used interchangeably. 

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra


checker/src/tag_domain.rs
```diff
@@ -17,12 +18,80 @@ macro_rules! does_propagate_tag {
     };
 }
 
+/// A replication of the `DefId` type from rustc. The type is used to implement serialization.
```
hermanventer, line 21: I have avoided serializing DefId structures because they are compilation specific and the intent of the serializer is to produce summaries that can be used across compiles. For example the summaries produced by running the compiler over the standard library are used by subsequent compilations on completely different environments.

However, performance considerations and limitations on concurrent access to the persistent summary store has led to a situation where the persistent summary store is only used for standard library summaries, which may be complete overkill.

I expect to remove the persistent summary store at some point. Meanwhile, the erroneous serialization of the def ids you introduce here will be OK as long as no summaries in the persistent store contain them.

checker/src/tag_domain.rs
```diff
@@ -17,12 +18,80 @@ macro_rules! does_propagate_tag {
...
+#[derive(Ord, PartialOrd, Eq, PartialEq, Copy, Clone, Hash)]
+pub struct DefIdDef {
```
hermanventer, line 23: Perhaps SerializableDefId?

checker/src/tag_domain.rs
stonebuddha: I see. Do you have an idea what could be a good representation of tag types in the analyzer?

hermanventer: You can stick with the DefId. Elsewhere ExpressionType is used, which is not ideal and certainly not appropriate for tags.

PR #543, opened 2020-06-19
Add an expression kind for unknown tag checks

stonebuddha:
## Description

This commit introduces a new expression `UnknownTagCheck` to resolve tag queries on function parameters. The mechanism is similar to unknown model fields. When checking whether the value at a path is attached with a tag or not, if the path is not tracked by the current environment and is rooted at a parameter, we create an `UnknownTagCheck` to postpone the tag query until we get more information about the parameters from the caller.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra


checker/src/abstract_value.rs
```diff
@@ -2884,6 +2890,35 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                            tag_presence.logical_not()
+                        };
+                        return result;
+                    }
+                }
+
+                // Cannot refine this tag check. Return again an unknown tag check.
+                AbstractValue::make_from(
+                    Expression::UnknownTagCheck {
+                        path: path.clone(),
```
hermanventer, line 2915: I think you ought to refine path.

checker/src/tag_domain.rs
```diff
@@ -97,20 +97,35 @@ pub struct Tag {
...
 /// An element of the tag domain is essentially an over-approximation for present and absent tags.
 /// The approximation is denoted as a map from tags to lifted Boolean values (`BoolDomain`).
+/// Unrecorded tags are mapped to the `default` field of the tag domain element.
 /// If a tag is mapped to `True`, then it must be present.
 /// If a tag is mapped to `False', then it must be absent.
 /// If a tag is mapped to `Top`, then it may or may not be present.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Clone, Serialize, Deserialize)]
 pub struct TagDomain {
     map: RedBlackTreeMap<Tag, BoolDomain>,
+    default: BoolDomain,
```
hermanventer, line 107: I'm none too sure about this. Why not just add a (tag, top) entry to map?

checker/src/tag_domain.rs
```diff
@@ -121,13 +136,14 @@ impl TagDomain {
...
             map: self.map.insert(tag, val),
+            default: self.default,
         }
     }
 
     /// Return a lifted Boolean that indicates the presence of `tag` in the tag domain element.
     #[logfn_inputs(TRACE)]
     pub fn has_tag(&self, tag: &Tag) -> BoolDomain {
-        *self.map.get(tag).unwrap_or(&BoolDomain::False)
+        *self.map.get(tag).unwrap_or(&self.default)
```
hermanventer, line 146: I expect to see BoolDomain::Top here.

checker/src/tag_domain.rs
```diff
@@ -139,7 +155,10 @@ impl TagDomain {
             let new_val = cur_val.or(val);
             new_map.insert_mut(*tag, new_val);
         }
-        TagDomain { map: new_map }
+        TagDomain {
+            map: new_map,
+            default: self.default.or(&other.default),
```
hermanventer, line 160: This makes no sense to me.

checker/src/abstract_value.rs
```diff
@@ -2884,6 +2885,31 @@ impl AbstractValueTrait for Rc<AbstractValue> {
                     default.clone()
                 }
             }
+            Expression::UnknownTagCheck {
+                path,
+                tag,
+                checking_presence,
+            } => {
+                if let Some(val) = environment.value_at(&path) {
+                    let tag_presence = val.has_tag(tag);
```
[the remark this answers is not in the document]
stonebuddha: The logic is implemented by the callers of has_tag. I'm thinking about replacing BoolDomain with AbstractValue in tag domain elements for a better control of tag checking.

checker/src/tag_domain.rs
stonebuddha: If a tag is never attached to a value, the tap map will not have an entry for the tag. This is the reason why I put BoolDomain::False previously.

But the change in CallVisitor::handle_has_tag introduces a problem. Previously if a path is not in the environment, we construct an unknown tag check. Now the use of `lookup_path_and_refine_result` may return a variable, whose tag map should indeed be TOP.

Therefore, I want get_tags to return a tag domain element that maps every tag to TOP. To achieve that, I introduce the default field in TagDomain.

hermanventer: I think this is too blunt a solution. A given value may have some tags that are known and some that are not yet known.

checker/src/tag_domain.rs
stonebuddha: The default field is used to answer tag queries. If a tag is kept track in the map, we just look it up in the map; otherwise, we use the default as the check result.

checker/src/abstract_value.rs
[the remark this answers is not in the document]
hermanventer: UnknownTagCheck is an expression that refines to either a compile time constant expression or otherwise remains an UnknownTagCheck. By the time this refinement happens, the call to has_tag has long gone and the caller has nothing more to do.

checker/src/tag_domain.rs
stonebuddha: Yes, so the default field is only used to control tags that are not yet known. But there are two different situations:
- A local variable should initially have no tags.
- A function parameter should mark the presence of every tag as "not yet known".

checker/src/abstract_value.rs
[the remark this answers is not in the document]
stonebuddha: By "callers of has_tag" I also mean the refinement routine itself.

checker/src/tag_domain.rs
hermanventer: Indeed, but the knowledge of how to deal with this obviously lies with the caller of has_tag. You could arguably return Bottom here (which is what default seems to be), but I would prefer Top.

stonebuddha: Returning either Bottom or Top will presumably fail the following test:
``` rust
let non_secret = 12345;
verify!(does_not_have_tag(&non_secret, Taint));
```

hermanventer: Perhaps has_tag should return a bool?

hermanventer: It seems that the problem is that we don't known when we don't know.

hermanventer: Since tags can never be deleted, the absence of a tag always means no for a local and unknown for a parameter. So the only thing the caller needs to know is "is the tag present?".

stonebuddha: Well, it might work if has_tag just returns if the tag map contains the entry or not, and all the callers of has_tag do their job to distinguish false and unknown.

stonebuddha: I guess the issue here is that whether has_tag can already decide the result or not. If has_tag indeed needs information from the callers, we should let the callers make the decision.

hermanventer: I suspect that will make things easier to understand. Right now, clearly, things are not particularly easy to understand,

stonebuddha: I'm not sure whether the caller can easily figure things out. Consider the program below:
``` rust
let temp = local + param;
verify!(has_tag!(&temp, Taint));
```
If the tag sets for both `local` and `param` are empty, the function call to `has_tag` returns "tag is not present".
Then the caller would like to figure out whether the tag is absent or unknown.
But this depends on the propagation behavior of Taint:
- If the addition `+` blocks Taint, then temp does not have Taint.
- Otherwise, if the addition `+` propagates Taint, then whether temp has Taint or not is unknown.

checker/src/abstract_value.rs
```diff
@@ -2565,8 +2565,14 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             | Expression::HeapBlockLayout { .. }
             | Expression::Reference { .. }
             | Expression::UnknownModelField { .. }
-            | Expression::Variable { .. } => {
-                return TagDomain::for_empty_set();
+            | Expression::UnknownTagCheck { .. } => return TagDomain::empty_set(),
```
hermanventer, line 2568: The underlying value may actually have tags, just not the one this check is looking for. Better return those.

checker/src/abstract_value.rs
```diff
@@ -2565,8 +2565,14 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             | Expression::HeapBlockLayout { .. }
             | Expression::Reference { .. }
```
hermanventer, line 2566: If the referenced value has been tagged, we would like to see here.

checker/src/abstract_value.rs
```diff
@@ -2884,6 +2890,35 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
             }
+            Expression::UnknownTagCheck {
+                path,
+                tag,
+                checking_presence,
+            } => {
+                if let Some(val) = environment.value_at(&path) {
+                    let tag_presence = val.has_tag(tag);
+                    if tag_presence.is_top() {
+                        // Still unknown, fall through.
```
hermanventer, line 2901: Instead of falling through you could just put the UnknownTagCheck expression here.

checker/src/call_visitor.rs
```diff
@@ -1190,48 +1185,41 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
             let source_path = Path::new_deref(self.actual_args[0].0.clone())
                 .refine_paths(&self.block_visitor.bv.current_environment);
             trace!(
-                "MiraiHasTag: checking if {:?} has {}been attached with <{:?},{:?}>",
+                "MiraiHasTag: checking if {:?} has {}been attached with {:?}",
```
hermanventer, line 1188: been attached with -> been tagged with

checker/src/call_visitor.rs
```diff
@@ -1190,48 +1185,41 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+                .bv
+                .type_visitor
+                .get_path_rustc_type(&source_path, self.block_visitor.bv.current_span);
+            let source_value = self
+                .block_visitor
+                .bv
+                .lookup_path_and_refine_result(source_path.clone(), source_rustc_type);
+
+            // Decide the result of has_tag! or does_not_have_tag!.
+            let tag_presence = source_value.has_tag(&tag);
```
hermanventer, line 1206: I suspect that things will be more understandable if we had an additional method "does_not_have_tag".

checker/src/abstract_value.rs
stonebuddha: This depends on "whether a tag check can propagate other tags or not". What's your opinion here?  Should we make this controllable via tag propagation set?

checker/src/abstract_value.rs
stonebuddha: Similar to the issue on tag checks: can reference propagate tags?

checker/src/abstract_value.rs
stonebuddha: In this block there are two cases where we return UnknownTagCheck: either path is not in the environment, or tag_presence is top. Should we duplicate the creation of the UnknownTagCheck expression?

checker/src/abstract_value.rs
hermanventer: I can't quite see why I would like a reference to hide the tags of the value being referred to.

checker/src/abstract_value.rs
hermanventer: Thinking about this some more, I can't convince myself that a compile time only operation need to be taint checked. So, I think I now agree with the code as you wrote it.

checker/src/abstract_value.rs
hermanventer: Missed that. Leave it as is then.

checker/src/abstract_value.rs
stonebuddha: My assumption there was that heap addresses do not lack information. We can definitely make references propagate everything, which is safe and makes tag checks easier.

stonebuddha: Well, this does not really make tag checks easier. If references propagate tags, then dereferences should also propagate them for the same reason. If this is the case, then we might need to check `&&&&&...secret` when we just want to check `secret`.

hermanventer: Perhaps this is a non issue. Whenever we check for the presence or absence of a tag, the reference will get dereferenced before the tags domain is accessed.

stonebuddha:
I'm going to close this PR and submit a new one after I make things more clear.

PR #546, opened 2020-06-24
Support unknown tag checks

stonebuddha:
## Description

This commit introduces a new expression `UnknownTagCheck` to resolve tag queries on function parameters. The mechanism is similar to unknown model fields. When checking whether the value at a path is attached with a tag or not, if the path is not tracked by the current environment (e.g., it is rooted at a function parameter), we create an `UnknownTagCheck` to postpone the tag query until we get more information from the caller.

This commit also updates the documentation for the tag domain and renames several tag-related APIs. The documentation for `TagDomain` now has a more detailed explanation of the design choices, and the names of tag-related APIs are more consistent with their intended usage.

Other changes in this commits are mostly code cleanup for the tag adding/checking logic. Now MIRAI will raise an error on add_tag!/has_tag!/does_not_have_tag! macros if (i) the tag type is not a generic type whose first parameter is a constant of type `TagPropagationSet`, or (ii) the tagged value is a reference.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra


checker/src/call_visitor.rs
```diff
@@ -2108,7 +2119,9 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
                 match rustc_gen_args[1].unpack() {
                     GenericArgKind::Type(rustc_type) => tag_rustc_type = rustc_type,
                     _ => {
-                        assume_unreachable!("expected the second generic argument of tag-related function calls to be a type");
+                        assume_unreachable!(
```
hermanventer, line 2122: Assumptions should be motivated in a comment. WHY is it reasonable to expect the second generic argument to be a type? Where is this checked?

stonebuddha: These assumptions are documented by this function's documentation and also a comment on line 2117. This is ensured by the rust type checker: the second argument of add_tag!/has_tag! macros must be a type.

PR #552, opened 2020-06-26
Keep track of tag checks from preconditions

stonebuddha:
## Description

The commit improves the precision of tag checks by refining tag abstractions when refining an abstract value with respect to a path condition. For example, if the path condition assumes (e.g., by preconditions) `has_tag!(param, Secret)` where `param` is a parameter, we want to update the abstract value of `param` to reflect that it has tag `Secret`. The information is needed to propagate tag `Secret` precisely when `param` is used in a compound expression later in the function.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra


hermanventer:
Perhaps we should talk about this in person. I still want to know why a separate visitor is the right solution.

stonebuddha:
I'm going to close this PR but investigate this issue later after implementing SMT-based tag tracking.

PR #553, opened 2020-06-26
Factor out target pattern expansion

hermanventer:
## Description

Factor out the code in copy_or_move_elements that deals with expanding target path patterns and weak updates. Generalize it so that it can be reused for attaching tags.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/body_visitor.rs
+++ b/checker/src/body_visitor.rs
@@ -1386,16 +1386,12 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
         // a collection of values obtained from the qualifier. We need to copy/move those
         // individually, hence we use a helper to call copy_or_move_elements recursively on
         // each source_path expansion.
-        let expanded_source_pattern = self.expanded_source_pattern(
+        let expanded_source_pattern = self.try_expand_source_pattern(
             &target_path,
             &source_path,
-            |_self, target_path, expanded_path| {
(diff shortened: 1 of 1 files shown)


checker/src/body_visitor.rs
```diff
@@ -1386,88 +1386,44 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
...
-            );
+        // If the pattern is a known size, copy_of_move_elements is called recursively
+        // on each expansion of the target pattern and try_expand_target_pattern returns true.
+        // If not, all of the paths that might match the pattern are weakly updated to account
+        // for the possibility that this assignment might update them (or not).
+        let expanded_target_pattern = self.try_expand_target_pattern(
+            &target_path,
+            &source_path,
+            root_rustc_type,
+            |_self, target_path, source_path, root_rustc_type| {
```
stonebuddha, line 1410: The name `root_rustc_type` here might be a bit misleading. Maybe `elem_rustc_type`?

hermanventer: It is a bit subtle, but from the context of the call to _self.copy_or_move_elements, the parameter named root_rutstc_type is the type of whatever kind of value is rooted at source_path, so elem_rustc_type will be misleading as well. Also, when target pattern is expanded to include conditional address expressions, it will not even be true that recursive calls will be to elements of an array.

PR #555, opened 2020-06-30
Encode tag checks as SMT queries

stonebuddha:
## Description

This commit implements Z3 encoding for tag checks. To ensure that the expression domain is precise enough to reduce tag checks to SMT queries, `UnknownTagCheck` expressions now record values instead of paths. On the Z3 side, I declare a top-level logical predicate `has_tag` in a way that `has_tag(path, tag)` encodes an uninterpreted Boolean-valued function call that checks whether `path` is attached with `tag` or not. When MIRAI generates a Z3 AST for `UnknownTagCheck { operand, tag, .. }`, it invokes the `general_has_tag(operand, tag)` routine, which recursively synthesizes a tag-check query from sub-expressions. This routine works similar to `AbstractValue::get_tags`.

This commit also includes several tests that indicate bugs or loss of precision. One is that when we tag a function parameter that is a struct, we don't know paths rooted at it so we don't tag its fields. The other is that the current methodology for `copy_or_move_elements` and `attach_tag_to_elements` only tries to split a path after enumerating paths rooted at the target path, so updating `if ... { local1 } else { local2 }` would not update `local1.field` or `local2.field`.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -255,6 +255,35 @@ impl AbstractValue {
         )
     }
 
+    /// Returns a tag check on `operand`. If we can decide the presence/absence of tag, return
+    /// TRUE/FALSE. Otherwise, returns an unknown tag check.
+    #[logfn_inputs(TRACE)]
+    pub fn make_tag_check(
+        operand: Rc<AbstractValue>,
+        tag: Tag,
+        checking_presence: bool,
+    ) -> Rc<AbstractValue> {
+        let check_value = if checking_presence {
+            operand.has_tag(&tag)
+        } else {
+            operand.does_not_have_tag(&tag)
+        };
+        if check_value.is_top() {
+            // Cannot refine this tag check. Return again an unknown tag check.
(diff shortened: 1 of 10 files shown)


hermanventer:
The test failures are hard to make out. It seems like Z3 is crashing. Does this work on your local machine?

stonebuddha:
> The test failures are hard to make out. It seems like Z3 is crashing. Does this work on your local machine?

Yes, it works normally on my machine.

stonebuddha:
> Need to figure out CI test failures.

I reproduce the failure locally by using the z3 binary in the repo.

hermanventer:
Perhaps time to update the binary in the repo, then.

stonebuddha:
Updated Z3 version to 4.8.8.

huitseeker:
@hermanventer   When could we get a new MIRAI release to test this one ? :D

hermanventer:
No need to wait for a release. The Libra CI picks up the tip of master for every run.

PR #556, opened 2020-07-01
Adapt to new nightly rustc

hermanventer:
## Description

Pick up new rustc nightly.

## Type of change

## How Has This Been Tested?
./validate.sh (MIRAI on MIRAI now fails)
ran MIRAI over Libra


stonebuddha:
I encountered the following error when running validate.sh:
```
RUSTFLAGS="-Z always_encode_mir" RUSTC_WRAPPER=mirai RUST_BACKTRACE=1 MIRAI_LOG=warn cargo build --lib -p mirai
...
Error: Sort mismatch at argument #1 for function (declare-fun < (Int Int) Bool) supplied sort is <null>
```
It seems to be caused by Z3.

hermanventer:
Yes, this is noted in the test plan. It is not a land blocker. I'll look into it for a separate PR.

PR #562, opened 2020-07-07
Refine rather than simplify to reduce &*ptr to just ptr

hermanventer:
## Description

Doing the &*ptr -> ptr transformation inside the factory function for qualified paths makes it difficult to defer the refinement when needed to ensure that &* is treated differently from & composed with *.

Also fix is_slice_pointer to report that pointers to slice pointers are themselves slice pointers.

Finally, add a bunch of comments to document these intricacies a bit better.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/block_visitor.rs
+++ b/checker/src/block_visitor.rs
@@ -1217,18 +1217,32 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
                 ..
             } if *selector.as_ref() == PathSelector::Deref => {
                 // We are taking a reference to the result of a deref. This is a bit awkward.
-                // The deref essentially does a copy of the value denoted by the qualifier.
-                // If this value is structured and not heap allocated, the copy must be done
-                // with copy_or_move_elements.
+                // The deref, by itself, essentially does a copy of the value denoted by the qualifier.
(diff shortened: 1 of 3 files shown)


checker/src/block_visitor.rs
```diff
@@ -1217,18 +1217,32 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                    // so there is no type information to use to come up with the length part of
+                    // the fat pointer. This implies that the source value must itself be a fat
+                    // pointer. This is, however, not what we'll see if we get the type of the
+                    // qualifier path. The reason for that is the ambiguous semantics of the
+                    // deref operator: Mostly, doing a deref of a fat pointer means doing a
+                    // deref of the thin pointer since that is what we expect when de-referencing
+                    // a pointer, whether it is fat or thin.
+                    //
+                    // When combined with an address_of operator, however, deref just means to
+                    // copy the pointer, whether it is fat of thin. (In this case, it must be fat.)
```
stonebuddha, line 1241: whether it is fat of thin -> no matter that it is fat or thin

PR #570, opened 2020-07-14
Do not project all union fields onto field 0.

hermanventer:
## Description

Doing such a projection loses type information which we may need later to accurately copy structures. Fix this by always updating all fields of the union together.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


stonebuddha:
Is there a (new) test case for projections of union fields?

PR #572, opened 2020-07-15
Support verification of constant-time security

stonebuddha:
## Description

This commit adds a command-line option `--constant_time TYPE_NAME` where `TYPE_NAME` is the name of a top-level crate tag type (#532). When the option is specified, MIRAI uses the tag domain to reason about constant-time security: every branch condition should *not* have the tag specified by `TYPE_NAME`.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/annotations/src/lib.rs
+++ b/annotations/src/lib.rs
@@ -65,13 +65,49 @@ pub enum TagPropagation {
     UninterpretedCall,
 }
 
-/// Provide a way to create tag transfer masks. It is equivalent to bitwise-or of all its arguments.
+/// Provide a way to create tag propagation sets. It is equivalent to bitwise-or of all its arguments.
 #[macro_export]
 macro_rules! tag_propagation_set {
     ($($x:expr),*) => {
         0 $(| (1 << ($x as u8)))*
     };
 }
+
+/// A tag propagation set indicating a tag is propagated by all operations.
+pub const TAG_PROPAGATION_ALL: TagPropagationSet = tag_propagation_set!(
+    TagPropagation::Add,
+    TagPropagation::AddOverflows,
+    TagPropagation::And,
+    TagPropagation::BitAnd,
+    TagPropagation::BitNot,
+    TagPropagation::BitOr,
+    TagPropagation::BitXor,
+    TagPropagation::Cast,
+    TagPropagation::Div,
+    TagPropagation::Equals,
+    TagPropagation::GreaterOrEqual,
+    TagPropagation::GreaterThan,
+    TagPropagation::IntrinsicBinary,
+    TagPropagation::IntrinsicBitVectorUnary,
+    TagPropagation::IntrinsicFloatingPointUnary,
+    TagPropagation::LessOrEqual,
+    TagPropagation::LessThan,
+    TagPropagation::LogicalNot,
+    TagPropagation::Mul,
+    TagPropagation::MulOverflows,
+    TagPropagation::Ne,
+    TagPropagation::Neg,
+    TagPropagation::Or,
+    TagPropagation::Offset,
+    TagPropagation::Rem,
+    TagPropagation::Shl,
+    TagPropagation::ShlOverflows,
+    TagPropagation::Shr,
+    TagPropagation::ShrOverflows,
+    TagPropagation::Sub,
+    TagPropagation::SubOverflows,
+    TagPropagation::UninterpretedCall
+);
 
 /// Equivalent to a no op when used with an unmodified Rust compiler.
 /// When compiled with MIRAI, this causes MIRAI to associate (tag) the value with the given type.
--- a/checker/src/abstract_value.rs
+++ b/checker/src/abstract_value.rs
@@ -1626,9 +1626,6 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             Expression::Reference(path) | Expression::Variable { path, .. } => {
                 path.is_rooted_by_zeroed_heap_block()
             }
-            Expression::UnknownTagCheck { operand, .. } => {
-                operand.is_contained_in_zeroed_heap_block()
-            }
             _ => false,
         }
     }
--- a/checker/src/block_visitor.rs
+++ b/checker/src/block_visitor.rs
@@ -17,6 +17,7 @@ use crate::path::PathRefinement;
 use crate::path::{Path, PathEnum, PathSelector};
 use crate::smt_solver::SmtResult;
 use crate::summaries::{Precondition, Summary};
+use crate::tag_domain::Tag;
 use crate::utils;
 use crate::{abstract_value, type_visitor};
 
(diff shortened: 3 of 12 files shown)


checker/src/abstract_value.rs
```diff
@@ -1186,7 +1186,7 @@ impl AbstractValueTrait for Rc<AbstractValue> {
                 default.try_to_retype_as(target_type),
             ),
             Expression::TaggedExpression { operand, tag } => {
-                operand.try_to_retype_as(target_type).add_tag(*tag)
+                operand.try_to_retype_as(target_type).add_tag(tag.clone())
```
hermanventer, line 1189: clone is normally an indication of an expensive operation. Perhaps not appropriate here?

checker/src/abstract_value.rs
```diff
@@ -1626,9 +1626,6 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             Expression::Reference(path) | Expression::Variable { path, .. } => {
                 path.is_rooted_by_zeroed_heap_block()
             }
-            Expression::UnknownTagCheck { operand, .. } => {
```
hermanventer, line 1629: I'm left wondering why this is no longer needed.

checker/src/call_visitor.rs
```diff
@@ -2130,7 +2142,17 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
                     }
                 }
 
-                // The tag type should be a generic ADT whose first parameter is a constant.
+                // The tag type could be the pre-defined ConstantTimeTaint type in mirai_annotations.
+                if let TyKind::Adt(adt_def, ..) = tag_rustc_type.kind {
+                    if let "mirai_annotations::ConstantTimeTaint" =
```
hermanventer, line 2147: This is an expensive operation.

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,72 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
         assume!(targets.len() == values.len() + 1);
         let mut default_exit_condition = self.bv.current_environment.entry_condition.clone();
         let discr = self.visit_operand(discr);
+
+        // Check if the discriminant is attached with the ConstantTimeTaint tag.
+        // If so, report an error concerning constant-time security.
+        if self.bv.check_for_errors && self.bv.check_for_constant_time_security {
```
hermanventer, line 274: This probably should be factored out into a separate method.

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,72 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                AbstractValue::make_tag_check(discr.clone(), Tag::ConstantTime, false);
+            let (tag_absence_as_bool, entry_cond_as_bool) =
+                self.check_condition_value_and_reachability(&tag_absence_value);
+
+            // We perform the tag check if the current block is possibly reachable.
+            if entry_cond_as_bool.unwrap_or(true) {
+                match tag_absence_as_bool {
+                    None => {
+                        // We cannot decide the result of the tag check.
+                        // In this case, report a warning if the analyzed function is public.
```
hermanventer, line 285: !(analysis_root <=> public)

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,72 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                            span,
+                            "the branch condition has a ConstantTimeTaint tag",
+                        );
+                        self.bv.emit_diagnostic(error);
+                    }
+
+                    _ => {}
+                }
+
+                // We premote the tag check as a precondition if is not always true.
```
hermanventer, line 312: if is -> if it is

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,72 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                if !tag_absence_as_bool.unwrap_or(false)
+                    && !tag_absence_value.expression.contains_local_variable()
```
hermanventer, line 314: The comment does not mention this important proviso.

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,72 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+
+            // We perform the tag check if the current block is possibly reachable.
+            if entry_cond_as_bool.unwrap_or(true) {
+                match tag_absence_as_bool {
+                    None => {
+                        // We cannot decide the result of the tag check.
+                        // In this case, report a warning if the analyzed function is public.
+                        if self.bv.function_being_analyzed_is_root()
+                            || self.bv.preconditions.len() >= k_limits::MAX_INFERRED_PRECONDITIONS
+                        {
```
hermanventer, line 288: You may want to warn in the case where the condition cannot be promoted because it contains a local variable.

checker/src/tag_domain.rs
```diff
@@ -8,92 +8,48 @@ use crate::bool_domain::BoolDomain;
...
+#[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Clone, Serialize, Deserialize, Hash)]
+pub enum Tag {
+    /// A tag for analyzing constant-time security. The type is pre-defined in mirai_annotations.
+    ConstantTime,
+
+    /// A user-defined tag is represented as a pair of its tag kind and its propagation set.
+    /// The tag kind is the path of the declared tag type, and the propagation set is associated to the
+    /// tag type as a const generic parameter.
+    Custom {
+        def_path_str: Rc<String>,
```
hermanventer, line 24: You should rather use the KnownNames enum here.

stonebuddha: The tag types are defined by users. I used to use DefId here.

checker/src/abstract_value.rs
stonebuddha: The tag type now contains Rc<String> so it cannot be copied. Anyway add_tag needs a copy/clone to record the tag in the tag domain.

checker/src/abstract_value.rs
stonebuddha: My understanding of `AbstractValue::is_contained_in_zeroed_heap_block` is that it checks if the value itself is contained in a zeroed heap block. It is not the case for unknown tag checks.

checker/src/call_visitor.rs
```diff
@@ -2137,7 +2137,17 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
                     }
                 }
 
-                // The tag type should be a generic ADT whose first parameter is a constant.
+                // The tag type could be the pre-defined ConstantTimeTaint type in mirai_annotations.
+                if let TyKind::Adt(adt_def, ..) = tag_rustc_type.kind {
+                    if self.block_visitor.bv.tcx.def_path_str(adt_def.did)
```
hermanventer, line 2142: This is expensive. Rather use KnownNames for this. Better yet, don't single out this type at all.

annotations/src/lib.rs
```diff
@@ -65,14 +65,53 @@ pub enum TagPropagation {
...
+    TagPropagation::ShlOverflows,
+    TagPropagation::Shr,
+    TagPropagation::ShrOverflows,
+    TagPropagation::Sub,
+    TagPropagation::SubOverflows,
+    TagPropagation::UninterpretedCall
+);
+
+/// A pre-defined tag for constant-time security. It is propagated by all operations.
+pub struct ConstantTimeTaint {}
```
hermanventer, line 113: I am surprised that this type gets special treatment. In general, I am not really comfortable with only this type leading to constant time diagnostics. Let's discuss some more over VC.

checker/src/tag_domain.rs
```diff
@@ -8,92 +8,48 @@ use crate::bool_domain::BoolDomain;
...
-            where
-                E: de::Error,
-            {
-                Ok(value)
-            }
+    /// Obtain `self`'s propagation set. The constant-time tag will be propagated by all operations.
+    #[logfn_inputs(TRACE)]
+    pub fn get_prop_set(&self) -> TagPropagationSet {
+        match self {
+            Tag::ConstantTime => TAG_PROPAGATION_ALL,
```
hermanventer, line 41: I'm not sure this is desirable. Some operations may randomize secrets enough to amount to sanitization. This design is not flexible.

checker/src/block_visitor.rs
```diff
@@ -267,6 +268,35 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                        let span = self.bv.current_span;
+                        let error = self.bv.cv.session.struct_span_err(
+                            span,
+                            format!(
+                                "unknown tag type for constant-time verification: {}",
+                                tag_name
+                            )
+                            .as_str(),
+                        );
+                        self.bv.emit_diagnostic(error);
```
hermanventer, line 286: You only want to do this once per compilation.

checker/src/known_names.rs
```diff
@@ -128,8 +128,8 @@ impl KnownNamesCache {
             .or_insert_with(|| Self::get_known_name_for(tcx, def_id))
     }
 
-    /// Uses information obtained from tcx to figure out which well known name (if any)
-    /// this def id corresponds to.
+    /// Uses information obtained from tcx and the constant_time_tag_name field from command-line
+    /// options to figure out which well known name (if any) this def id corresponds to.
```
hermanventer, line 132: Please restore the comment.

checker/src/call_visitor.rs
```diff
@@ -2189,10 +2189,33 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
                 {
-                    Some(Tag {
+                    let tag = Tag {
                         def_id: tag_adt_def.did.into(),
                         prop_set: *data,
-                    })
+                    };
+
+                    // Check if tag is the constant-time verification tag specified by the user.
+                    if self.block_visitor.bv.cv.constant_time_tag_cache.is_none() {
```
hermanventer, line 2198: Please factor this out into a separate function.

PR #583, opened 2020-07-22
Reduce the size of the active calls map.

hermanventer:
## Description

Tweaks to make log output easier to read. Rewording some classical terms that no longer sound so nice.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/call_visitor.rs
+++ b/checker/src/call_visitor.rs
@@ -973,17 +973,19 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
                 trace!("post env {:?}", self.block_visitor.bv.current_environment);
                 return;
             } else {
-                let saved_callee_def_id = self.callee_def_id;
-                self.callee_def_id = def_id;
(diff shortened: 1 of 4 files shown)


checker/src/call_visitor.rs
```diff
@@ -973,10 +973,12 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
                 return;
             } else {
-                let saved_callee_def_id = self.callee_def_id;
-                self.callee_def_id = def_id;
-                self.deal_with_missing_summary();
-                self.callee_def_id = saved_callee_def_id;
+                if self.block_visitor.bv.check_for_errors {
+                    let saved_callee_def_id = self.callee_def_id;
+                    self.callee_def_id = def_id;
+                    self.deal_with_missing_summary();
```
stonebuddha, line 979: On line 988, there is another `deal_with_missing_summary` call that is not guarded by `check_for_errors`.

PR #584, opened 2020-07-23
Add a tutorial for using tag analysis with MIRAI

stonebuddha:
## Description

This commit adds a tutorial for tag analysis, which includes an overview of tag analysis in MIRAI, the mechanisms for declaring and adding/checking tags, and three scenarios of using tag analysis. This commit also adds three examples that correspond to the three scenarios, respectively.

## Type of change

None of the above: documentation change with some new examples.

## How Has This Been Tested?

./validate.sh

### DIFF
--- a/documentation/TagAnalysis.md
+++ b/documentation/TagAnalysis.md
@@ -1,0 +1,227 @@
+# Tag Analysis
+
+## Overview
+
+Tag analysis is a compile-time static analysis that keeps track of information flows in programs. Information flows
+from variable X to variable Y, whenever information stored in X is transferred to Y. Tag analysis attaches different
+kinds of tags to values in a program to keep track of data flows for these tags. For example, taint analysis is a
+special form of tag analysis that reasons about sensitive information, such as private user data in social network
+service, or private keys in crypto code. Using tag analysis, a static analyzer can verify that tainted values (i.e.,
+values attached with special taint tags) will never flow to specific program locations that may leak the information,
+e.g., a method writing the tainted values to a public channel. Because the analysis is carried out at compile time,
+the analyzer can verify non-trivial information-flow properties without any runtime overheads.
+
+MIRAI supports a context-, field-, flow-, and path-sensitive tag analysis. Context-sensitivity and field-sensitivity
+come from MIRAI's summary-based inter-procedural analysis and precise abstract heap model, respectively.
+We implement the tag analysis as a data-flow analysis with a tag domain that over-approximates present and absent tags
+on values simultaneously. When the data-flow analysis is not precise enough to answer queries, MIRAI generates a
+propositional formula that encodes both the query and the path condition that guards the query, and uses an
+off-the-shelf constraint solver such as [Z3](https://github.com/Z3Prover/z3) to answer the query.
+
+Tag analysis is in general an undecidable problem, so MIRAI might not get sufficiently precise information to verify
+information-flow properties. MIRAI's tag analysis is developed to be consistent with MIRAI's design, which avoids false
+negatives (not flagging real errors), and reduces the number of false positives (flagging false errors). When the
+analysis result is not precise enough, or MIRAI timeouts because the analyzed program has very complex control or data
+flow, MIRAI will produce a conservative approximation of the program states and report possible errors, some of which
+may be false positives. As in other static analysis, developer-provided annotations, such as loop invariants, or
+simplification of the control flow in the analyzed program, could help MIRAI improve analysis precision.
+
+## Declaring tags
+
+MIRAI supports developer-defined tags and customizable tag propagation behavior. Tag-related types and macros are
+provided by the [MIRAI Annotations](https://crates.io/crates/mirai-annotations) crate.
+
+In MIRAI, tags are just Rust types, except that they are assumed to have at least one generic argument, the first of
+which should be a const parameter of type `TagPropagationSet`. The code below declares a tag kind named
+`SecretTaintKind`.
+
+``` rust
+#![feature(const_generics)]
+#![allow(incomplete_features)]
+
+#[macro_use]
+extern crate mirai_annotations;
+
+use mirai_annotations::{TagPropagation, TagPropagationSet};
+
+struct SecretTaintKind<const MASK: TagPropagationSet> {}
+```
+
+`MASK` is used to specify propagation behavior of different program operations for the tag. For example, if bitwise-xor
+can be used to “sanitize” tainted data, it will then block the taint tag. MIRAI provides a macro `tag_propagation_set!`
+to create a tag-propagation set by specifying all the operations that can propagate the tag. The code below defines a
+tag named `SecretTaint` that can only be propagated by equality checks.
+
+``` rust
+const SECRET_TAINT_MASK = tag_propagation_set!(TagPropagation::Equals, TagPropagation::Ne);
+
+type SecretTaint = SecretTaintKind<SECRET_TAINT_MASK>;
+```
+
+## Attaching and checking tags on values
+
+In MIRAI, tags are attached to values via the `add_tag!` macro, whose first parameter is a reference to the value
+that is being tagged, and second parameter is a well-defined tag type. The code below attaches the `SecretTaint` tag we
+just defined to a value.
+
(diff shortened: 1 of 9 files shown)


examples/tag_analysis/timing_channels/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+type SecretTaint = SecretTaintKind<SECRET_TAINT_MASK>;
+
+pub mod non_constant_time {
+    /// A compare function that is **not** constant-time.
+    pub fn compare(secret: &[i32], public: &[i32], len: usize) -> bool {
+        precondition!(secret.len() == len);
+        precondition!(public.len() == len);
+        let mut i = 0;
+        while i < len {
+            assume!(has_tag!(&secret[i], crate::SecretTaint));
```
hermanventer, line 32: The assumption here amounts to precondition, so rather use a precondition.

In general, though, we should not expect code to have such preconditions. Rather, the summary of this function will contain a conditional expression where the condition, after refinement, will be tainted. That is the place to give the diagnostic.

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
+# Tag Analysis
+
+## Overview
+
+Tag analysis is a compile-time static analysis that keeps track of information flows in programs. Information is said
```
huitseeker, line 5: "Information flows" would avoid the passive

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+to flow from object X to object Y, whenever information stored in X is transferred to object Y. Tag analysis attaches
+different kinds of tags to objects in a program to keep track of dataflows for these tags. For example, taint analysis
```
huitseeker, line 7: data flows"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+is a special form of tag analysis that reasons about sensitive information, such as private user data in social network
+service, or private keys in crypto code. Using tag analysis, a static analyzer can verify that tainted objects (i.e.,
+objects attached with special taint tags) will never flow to specific program locations that may leak the information,
+e.g., a method writing the tainted objects to a public channel. Because the analysis is carried out at compile time,
+the analyzer can verify non-trivial information-flow properties without any runtime overheads.
+
+The tag analysis provided by MIRAI is context-, field-, flow-, and path-sensitive. Context-sensitivity and field-
+sensitivity come from MIRAI's summary-based inter-procedural analysis and precise abstract heap model, respectively.
```
huitseeker, line 15: "sensitivity comes"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+The tag analysis is implemented as a dataflow analysis with a tag domain that over-approximates present and absent tags
```
huitseeker, line 16: "data flow"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+on program objects simultaneously. When the dataflow analysis is not precise enough to answer queries, MIRAI generates
+a propositional formula that encodes both the query and the path condition that guards the query, and uses an
+off-the-shelf constraint solver such as [Z3](https://github.com/Z3Prover/z3) to answer the query.
+
+Tag analysis is in general an undecidable problem, so it might not obtain sufficiently precise information to verify
```
huitseeker, line 21: "it might not get"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+information-flow properties. The tag analysis provided by MIRAI is developed to consist with MIRAI's design, which
+avoids false negatives (not flagging real errors), and reduces the number of false positives (flagging false errors).
```
huitseeker, line 23: Since this point seems particularly important, I would love to read a short explanation of the explicit link between this, positive/negative taint analysis, and timeouts.
I think you already have the main elements for that (i.e. examples where you check presence/absence of a tag). What's missing, I think, is one sentence here on how timeouts might occur when the tracking blows up, and once sentence per example below laying out the "in situ" consequence: "here, if the data flow was *really* complex, MIRAI would be unable to check for the presence of the tag and would report an error"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+
+See the [timing-channels crate](../examples/tag_analysis/timing_channels) for a complete example.
+
+## Scenario II: Track untrustworthy inputs
+
+In public-key cryptography, the source of a public key might be untrustworthy. Also, public keys obtained from the
+the outside environment might suffer from non-trivial vulnerabilities, e.g., public keys for Ed25519 might not be safe
+against small subgroup attacks.
+
+The code below illustrates a scenario where we want to ensure that every public key used in the system is obtained by
```
huitseeker, line 141: "we want to make sure"

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,187 @@
...
+the `try_from` method that checks for key validity. We can verify this property using tag analysis: whenever a public
```
huitseeker, line 142: "when" a public key

examples/tag_analysis/timing_channels/src/lib.rs
stonebuddha: Removed the assumption, but added a precondition about the tag on `len`. The precondition is supposed to be promoted from `!UnknownTagCheck([[i < len]], SecretTaint)`, but the promotion is currently not supported because the condition contains a local variable `i`.

hermanventer: Meanwhile, put the precondition on the entire array and rely on TagPropagation::SubComponent to propagate the tag to the individual elements. (That too, might now work right now, so make an issue if need be.)

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,206 @@
+# Tag Analysis
+
+## Overview
+
+Tag analysis is a compile-time static analysis that keeps track of information flows in programs. Information flows
+from value X to value Y, whenever information stored in X is transferred to value Y. Tag analysis attaches different
```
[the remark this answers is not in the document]
stonebuddha: When we add a tag to a variable, the tag is attached to the value bound by the variable. I think "object" could be a proper name, but that might be confusing because Rust is not object-oriented.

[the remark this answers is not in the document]
hermanventer: The terms are a bit loose. In this context there is no difference between information and values. If you can store something in X, then X is a place/location/variable, not a value or information, the latter being the thing that you store in the place/location/variable.

documentation/TagAnalysis.md
```diff
@@ -0,0 +1,206 @@
...
+
+MIRAI's constant-time verification is enabled by a command-line option `--constant_time Name`, where `Name` is the tag
+kind type identifying sensitive information that should not influence running time, e.g., the type `SecretTaintKind`
+defined above. MIRAI reports every branch condition that has the tag used for constant-time verification. If MIRAI
+timeouts or the analysis result is not precise enough, MIRAI will report all possible errors, which may contain false
+positives.
+
+See the [timing-channels crate](../examples/tag_analysis/timing_channels) for a complete example.
+
+## Scenario II: Track untrustworthy inputs
```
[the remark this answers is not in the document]
huitseeker: I'm really attached to developers the different meanings of "false positives" vs "false negatives" in this various contexts like this, so while I agree the disjunction may be artificial, I believe we can afford to be a bit long-winded here. Plus, developers haven't really had to deal with an *overabundance* of ways to use taint analysis yet, on the contrary.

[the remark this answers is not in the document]
stonebuddha: Updated this scenario and added comments about false positives.

examples/tag_analysis/timing_channels/src/lib.rs
stonebuddha: When I tried to attach a tag to the array, I found it impossible to tag values with unsized types. Changed the parameters in the example to fixed-length arrays and submitted a PR #588 for the issue.

PR #585, opened 2020-07-23
Allow recursive calls after devirtualization

hermanventer:
## Description

Reorganize code so that devirtualization happens before the check for recursion.

Work around some hard to repro bugs that get uncovered by the additional analysis. Fix the way Z3 expressions are generated for switch statements with discriminator values that are bit vectors.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/call_visitor.rs
+++ b/checker/src/call_visitor.rs
@@ -129,8 +129,6 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
             self.block_visitor.bv.assume_function_is_angelic |= call_was_angelic;
             self.block_visitor.bv.start_instant = Instant::now() - elapsed_time;
             return summary;
-        } else if let Some(devirtualized_summary) = self.try_to_devirtualize() {
-            return devirtualized_summary;
         }
         Summary::default()
     }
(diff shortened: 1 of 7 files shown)


checker/src/call_visitor.rs
```diff
@@ -139,11 +137,19 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
     /// concrete method that implements the given virtual method and returns the summary of that,
     /// computing it if necessary.
```
stonebuddha, line 138: The method `try_to_devirtualize` now does not return a summary.

checker/tests/run-pass/generic_recursive.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
```
stonebuddha, line 6: Maybe a comment for what the code below  is testing?

checker/src/call_visitor.rs
hermanventer: Yes, it updates the state of the call visitor, or not.

PR #588, opened 2020-07-24
Allow tags on values of unsized types

stonebuddha:
## Description

Rust's generic type parameters are implicitly bound by `Sized`. This commit explicitly annotates the generic type parameter of tag-related functions with `?Sized` to allow tags on values of unsized types.

This commit fixes a bug in the implementation of add_tag/check_tag: when the source address is a fat pointer (e.g., the value being tagged is a slice), we obtain the correct thin pointer before doing the actual dereference.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/annotations/src/lib.rs
+++ b/annotations/src/lib.rs
@@ -1065,17 +1065,17 @@ pub fn mirai_abstract_value<T>(x: T) -> T {
 
 // Helper function for MIRAI. Should only be called via the add_tag! macro.
 #[doc(hidden)]
-pub fn mirai_add_tag<V, T>(_v: &V) {}
+pub fn mirai_add_tag<V: ?Sized, T>(_v: &V) {}
 
 // Helper function for MIRAI. Should only be called via the has_tag! macro.
 #[doc(hidden)]
-pub fn mirai_has_tag<V, T>(_v: &V) -> bool {
+pub fn mirai_has_tag<V: ?Sized, T>(_v: &V) -> bool {
     false
 }
 
 // Helper function for MIRAI. Should only be called via the does_not_have_tag! macro.
 #[doc(hidden)]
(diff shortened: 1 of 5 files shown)


annotations/src/lib.rs
```diff
@@ -1065,17 +1065,17 @@ pub fn mirai_abstract_value<T>(x: T) -> T {
 
 // Helper function for MIRAI. Should only be called via the add_tag! macro.
 #[doc(hidden)]
-pub fn mirai_add_tag<V, T>(_v: &V) {}
+pub fn mirai_add_tag<V: ?Sized, T>(_v: &V) {}
```
hermanventer, line 1068: There are more instances of this in the annotations crate. You may as well fix them too.

checker/src/call_visitor.rs
```diff
@@ -1007,7 +1007,20 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
         precondition!(self.actual_args.len() == 1);
 
         if let Some(tag) = self.extract_tag_kind_and_propagation_set() {
-            let source_path = Path::new_deref(self.actual_args[0].0.clone())
+            let source_pointer_path = self.actual_args[0].0.clone();
+            let source_pointer_rustc_type = self
+                .block_visitor
+                .bv
+                .type_visitor
+                .get_path_rustc_type(&source_pointer_path, self.block_visitor.bv.current_span);
```
hermanventer, line 1015: Instead of doing this expensive and brittle operation, you can just get the type from self.actual_argument_types.

annotations/src/lib.rs
stonebuddha: I had a look at other instances but they are different because an unsized type needs to be used behind some kind of pointer.

PR #591, opened 2020-07-27
Implement intrinsics::write_bytes

hermanventer:
## Description

Implement intrinsics::write_bytes.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


checker/src/call_visitor.rs
```diff
@@ -1754,6 +1758,92 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+            }
+            source_value
+        }
+        if elem_type.is_primitive() {
+            let dest_pattern = Path::new_slice(dest_path, count_value);
+            let source_value = repeated_bytes(elem_size, byte_value);
+            self.block_visitor
+                .bv
+                .current_environment
+                .update_value_at(dest_pattern, source_value);
```
stonebuddha, line 1794: Is it supposed to be a `copy_or_move_elements` call, because the target path is a pattern?

hermanventer: No, we don't want expansion of the pattern at this point since the caller is going to do that anyway.

stonebuddha: Ah, the caller will do that during refinement?

hermanventer: The caller's caller, yes.

PR #593, opened 2020-07-28
Some more refinements

hermanventer:
## Description

Some more refinements

## Type of change

## How Has This Been Tested?
./validate.sh (new test case will come in future PR)
ran MIRAI over Libra


stonebuddha:
Should `selector` in the two sites below be `refined_selector`?
https://github.com/facebookexperimental/MIRAI/blob/0836c862712c0035c57defb12b37eb335839e8a6/checker/src/path.rs#L756-L759
https://github.com/facebookexperimental/MIRAI/blob/0836c862712c0035c57defb12b37eb335839e8a6/checker/src/path.rs#L822-L830

PR #594, opened 2020-07-29
Attach tags to implicit tag fields

stonebuddha:
## Description

When a tag is attached to a non-scalar value (e.g., a struct), and it can be propagated to sub-components, we need to attach the tag to the tag field of the value's non-scalar sub-components. Tag fields of sub-components are sometimes not tracked in the environment. This commit implements a mechanism to update such implicit tag fields.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/body_visitor.rs
+++ b/checker/src/body_visitor.rs
@@ -2135,7 +2135,7 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
 
                 _self.non_patterned_copy_or_move_elements(
                     sub_path.clone(),
-                    sub_path,
+                    sub_path.clone(),
                     root_rustc_type,
                     false,
                     |_self, path, _, new_value| {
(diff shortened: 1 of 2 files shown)


checker/src/body_visitor.rs
```diff
@@ -2248,4 +2256,53 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
...
+
+                    if !path_prefix_rustc_type.is_scalar() {
+                        let (tag_field_path, tag_field_value) = self
+                            .extract_tag_field_of_non_scalar_value_at(
+                                path_prefix,
+                                path_prefix_rustc_type,
+                            );
+
+                        // Perform an update when tag_field_path is not tracked in the old environment.
+                        // In this case, tag_field_path represents an implicit tag field.
```
hermanventer, line 2297: It is not clear to me why you attach the tag only if the prefix does not already have a tag field. I would like an explanation and a test case to prove it is not needed.

stonebuddha: Those existing tag fields are already updated by non_patterned_copy_or_move. Added a comment and a test to clarify this.

PR #595, opened 2020-07-29
Check tags on the prefixes of a qualified path

stonebuddha:
## Description

When a function parameter has a non-scalar type (e.g., struct), MIRAI does not have enough information about possible paths rooted by the parameter. Thus, when we add a tag to the function parameter, and the tag can be propagated to sub-components, we need a mechanism to propagate the tag to possible paths rooted by the parameter.

This commit focuses on intra-procedural verification. When the program adds and checks a tag on a qualified path (rooted by a parameter) inside the same function, at the checking site we iterate over all the prefixes of the path to see whether it can be tagged or not. Tag propagation during refinements will be addressed in a future PR.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/call_visitor.rs
+++ b/checker/src/call_visitor.rs
@@ -1134,15 +1134,73 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
             } else {
                 self.block_visitor
                     .bv
-                    .lookup_path_and_refine_result(source_path, source_rustc_type)
+                    .lookup_path_and_refine_result(source_path.clone(), source_rustc_type)
             };
 
             // Decide the result of has_tag! or does_not_have_tag!.
-            result = Some(AbstractValue::make_tag_check(
-                tag_field_value,
-                tag,
-                checking_presence,
-            ));
+            let mut check_result =
+                AbstractValue::make_tag_check(tag_field_value, tag, checking_presence);
+
+            // If the tag can be propagated through sub-components, and source_path is rooted by a function parameter,
+            // we need to check the tag on the values that can contain source_path as a sub-component. Note that if
+            // source_path is rooted by a local variable, MIRAI has already propagated the tag to source_path when the
+            // tag was added to the value that contains source_path as a sub-component.
+            // Operationally, source_path is a qualified path and we check if any of its prefix has the tag (when checking_presence = true),
+            // or if all of its prefixes does not have the tag (when checking_presence = false).
+            if tag.is_propagated_by(TagPropagation::SubComponent)
+                && source_path.is_rooted_by_parameter()
+            {
+                let mut path_prefix = &source_path;
+                while let PathEnum::QualifiedPath { qualifier, .. } = &path_prefix.value {
+                    path_prefix = qualifier;
+
+                    let path_prefix_rustc_type = self
+                        .block_visitor
+                        .bv
+                        .type_visitor
+                        .get_path_rustc_type(&path_prefix, self.block_visitor.bv.current_span);
+                    if !path_prefix_rustc_type.is_scalar() {
+                        let tag_field_value = self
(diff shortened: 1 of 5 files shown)


checker/src/call_visitor.rs
```diff
@@ -1134,15 +1134,63 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+                let mut path_prefix = &source_path;
+                while let PathEnum::QualifiedPath { qualifier, .. } = &path_prefix.value {
+                    path_prefix = qualifier;
+
+                    let path_prefix_rustc_type = self
+                        .block_visitor
+                        .bv
+                        .type_visitor
+                        .get_path_rustc_type(&path_prefix, self.block_visitor.bv.current_span);
+                    if !path_prefix_rustc_type.is_scalar() {
```
hermanventer, line 1162: I'm none too sure that this check is needed. After all, source_path addresses a sub component of the value at qualifier.

checker/src/call_visitor.rs
```diff
@@ -1134,15 +1134,63 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+                            .extract_tag_field_of_non_scalar_value_at(
+                                &path_prefix,
+                                path_prefix_rustc_type,
+                            )
+                            .1;
+
+                        if checking_presence {
+                            // We are checking presence of a tag. It is equivalent to *any* prefix having the tag.
+                            // Thus we use a logical or.
+                            check_result = check_result.or(AbstractValue::make_tag_check(
```
hermanventer, line 1175: if tag_field_value is known to be true at this point, you can return true here.

checker/src/call_visitor.rs
```diff
@@ -1134,15 +1134,63 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
+                                tag_field_value,
+                                tag,
+                                checking_presence,
+                            ));
+                        } else {
+                            // We are checking absence of a tag. It is equivalent to *all* prefixes not having the tag.
+                            // Thus we use a logical and.
+                            check_result = check_result.and(AbstractValue::make_tag_check(
```
hermanventer, line 1183: If tag_field_value is known to be true at this point, you can return false here.

checker/src/call_visitor.rs
stonebuddha: My concern was that path selectors are not really sub-components; some of them are introduced by MIRAI and are not programmer-accessible. Thus I introduce a predicate `Path::is_an_accessible_component_of(root_path)` to check is a qualified path is a programmer-accessible sub-component of root_path.

hermanventer: I don't think that is necessary either. Attaching a tag field to a primitive value is not wrong as long as you don't depend on it being there in any copies.

stonebuddha: Yes, that's true. But I was thinking of the dereference selector, and it seems that tags of `a.b` should not be propagated to `a.b.deref`. In that sense, the deference selector should not be treated as a sub-component.

hermanventer: It probably won't hurt, but it does smell a bit odd. I guess I prefer your original formulation then.

stonebuddha: Reverted the changes. Two small tweaks: (i) Avoid adding tags to `Layout` and `Discriminant`: tags on those paths seem not useful, and in the codebase there are pattern matches on the values located at those paths, so wrapping the values with tags would be problematic. (ii) Let the `propagate_to_tag_fields` method handle both existing and implicit tag fields. The problem was that some tag field paths are not rooted by the source_path (e.g., the tag field path of `param.deref` is `param.$tag`).

PR #596, opened 2020-07-30
Make Option::take work

hermanventer:
## Description

Functions that modify mutable reference parameters and then also return the original values in their summaries are problematic because the side effects and original values use the same paths and this can cause the original values to resolve to updated values. The new test case is a case in point, because Option::take returns the old value of the option while also setting the option to None. This sets up a conflict at the call site between a reference to the old discriminator value of the option and the new one.

To solve this, it is necessary to identify when a reference to a parameter refers to its original value. This happens when the original value is copied into a local via a dereference operation (if the parameter value is not a mutable reference there can be no side-effect, so there is no issue). This is checked for on line 834 of path.rs.

Since the value of the copy is not known, it becomes a Variable with path pointing to the parameter. The new trick introduced by this PR is to add PathSelector::CopyOfOriginal to this path (and then ensure that it remains the last selector). Consequently, any path constructed from the copy will not resolve to any modification of the parameter and the path can be included into the method summary without creating ambiguity.

At the call site, path refinement that encounters PathSelector::CopyOfOriginal must then look up the qualifier path in the environment that pertained at the start of the call (the pre-state) rather than the current environment, which may already contain side-effects.

Much the unfortunate size of this PR comes from the need this creates to keep track of the pre-state.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


checker/src/abstract_value.rs
```diff
@@ -2982,7 +3012,19 @@ impl AbstractValueTrait for Rc<AbstractValue> {
...
+                        qualifier,
+                        selector,
+                        ..
+                    } = &refined_path.value
+                    {
+                        if *selector.as_ref() == PathSelector::CopyOfOriginal {
+                            if let Some(val) = pre_environment.value_at(qualifier) {
+                                return val.clone();
+                            }
+                        }
```
stonebuddha, line 3026: Should this routine be applied to UnknownModelField and UnknownTagField, because these two cases also contain an environment lookup?

checker/src/path.rs
```diff
@@ -814,7 +831,20 @@ impl PathRefinement for Rc<Path> {
...
+                                        // This is awkward since we now have to distinguish between the original value of the parameter
+                                        // and any subsequent modifications to the parameter. We use PathSelector::CopyOfOriginal for this purpose.
+                                        refined_path =
+                                            Path::new_copy_of_original_value(qualifier.clone());
+                                    } else {
+                                        refined_path = qualifier.clone();
+                                    }
                                 }
                             }
                             return Path::new_qualified(refined_path, refined_selector);
```
stonebuddha, line 850: When refined_path is obtained from line 844, i.e., it's a qualified path whose selector is CopyOfOriginal, the invariant that CopyOfOriginal is always the last selector does not hold.

hermanventer:
I want to rethink this.

PR #600, opened 2020-08-04
Keep track of parameter values that are structs

### LINKED ISSUE #597: Incomplete summary of side effects
## Issue

MIRAI seems to produce an incomplete summary of side effects, when the summarized function mutates a path rooted by a parameter X, and then copies the parameter X to another parameter Y.

## Steps to Reproduce

Run MIRAI on the following code:
``` rust
fn mutate_and_then_copy(dest: &mut (i32, i32), src: &mut (i32, i32)) {
    src.1 = 789;
    *dest = *src;
}

pub fn main() {
    let mut a = (123, 456);
    let mut b = (456, 123);
    mutate_and_then_copy(&mut a, &mut b);
    verify!(b.0 == 456 && b.1 == 789);
    verify!(a.0 == 456 && a.1 == 789);
}
```

## Expected Behavior

The program contains no bugs, so MIRAI is supposed to report no errors.

## Actual Results

MIRAI reports a provably false verification condition:
```
error: provably false verification condition
   --> src/main.rs:614:5
    |
614 |     verify!(a.0 == 456 && a.1 == 789);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

## Environment

rustc 1.46.0-nightly (2753fab7c 2020-07-05)

hermanventer:
## Description

When a function updates a structure via a mutable reference, with the collective value of another parameter, and then also updates an individual field of the mutable reference, the summary must include a side effect that is the collective path, effectively something like param_i.* = param_j.*, as well as a subsequent side effect for the individual field, for example: param_i.f = 123.

Because of the ambiguity (aliasing) between param_i.* for param_i.f, the side effects are sorted so that more inclusive copy operations precede less inclusive operations. Hence param_i.* = .. will precede param_i.f in the ordered list of side effects.

Fixes #597 

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


checker/src/body_visitor.rs
```diff
@@ -1821,6 +1821,15 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
         // be saddled with removing it. This case corresponds to no_children being true.
         if val_type == ExpressionType::NonPrimitive {
             // First look at paths that are rooted in rpath.
+            // Remove an explicit deref from source_path to make it a valid root.
```
stonebuddha, line 1824: This may also happen in other methods. Do we need to check all the sites where a path is used as a root?

hermanventer: I've checked all the sites and this is only place where it can happen.

PR #601, opened 2020-08-05
Distinguish between parameter values and copies of such values in locals

hermanventer:
## Description

When a parameter gets copied to a local, the value of the local is not known and it needs to refer back to the parameter so that call site refinement can deal with values obtained in whole or part from the local. Such references need to be distinct from the parameter itself, otherwise the local becomes an alias for the parameter and this is not semantically valid if the parameter is modified after the copy is made.

To deal with this, a new kind of path, PathEnum::ParameterCopy is introduced and it becomes the root of the path of any Expression::Variable where the path root refines to a PathEnum::Parameter. This can only happen in an environment where the path does not have a known value and it keeps the path distinct from subsequent modifications to the parameter, which will show up as paths rooted in the Parameter, not the ParameterCopy.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


checker/src/path.rs
```diff
@@ -649,6 +666,9 @@ pub trait PathRefinement: Sized {
     /// and not have more than one path for the same location.
     fn refine_paths(&self, environment: &Environment) -> Rc<Path>;
 
+    /// Returns a copy of self with a Parameter root replaced by a ParameterCopy root.
+    fn replace_parameter_root_with_copy(&self) -> Rc<Path>;
```
stonebuddha, line 670: Maybe add some comments about when this method should be applied to a path?

stonebuddha:
Thanks for the comments. If I understand it correctly, every site of `make_typed_unknown` may need to replace a parameter root with ParameterCopy.

https://github.com/facebookexperimental/MIRAI/blob/73c56c3674e774f7de37c4cd542275479efcb563/checker/src/body_visitor.rs#L1914-L1920

Is it necessary to also do such replacement here in weak updates?

hermanventer:
Paths used as keys in the environment refer to the current state, not the pre-state, so they should be kept as is. This is all rather subtle and I'll have to write it all up in some detail as part of free standing documentation. But for this PR, I don't think more comments will be all that helpful.

Issue #389, opened 2020-01-10
Simple assertions not failing as expected


zvonimir:
## Issue

I have been working on porting SMACK Rust regressions to MIRAI to uncover potential issues. The eventual goal is to have a nice publicly available suite of Rust benchmarks.
Along the way I discovered this MIRAI issues, where it appears to be missing the fact that the two assertions in this small example should obviously fail.

## Steps to Reproduce

Run MIRAI on this small example:
```
#[macro_use]
extern crate mirai_annotations;

pub fn main() {
  let a : i32 = result!();
  let b : i32 = result!();
  assume!(4 < a && a < 8); // a in [5,7]
  assume!(5 < b && b < 9); // b in [6,8]
  verify!(36 <= a * b); // a*b in [30,56]
  verify!(a * b <= 49); // a*b in [30,56]
}
```

## Expected Behavior

Both assertions should fail.

## Actual Results

The assertions are not failing.

## Environment

rustc 1.42.0-nightly (0de96d37f 2019-12-19)
MIRAI v1.0.5

hermanventer:
Both calls to result!() produce the same abstract value (local_0 in MIR), so the assumptions are contradictory and anything goes. This would be the test case for introducing a new way to specify an abstract value that is not meant to be the result of a function.

zvonimir:
Ah, I see. They are not quite contradictory since there is an overlap between the intervals they constrain. Bug this explains the weird outcomes I've been observing.

zvonimir:
I created a more general discussion to resolve this in #397. It would be good to reach some kind of an agreement there.
Related this this particular issue, I am used to functions like `result!` creating a fresh abstract value with every invocation. That is how verifiers that I am familiar with operate at least.

PR #608, opened 2020-08-08
Add warning if assumption is provably false (resolve #393)

### LINKED ISSUE #393: False assumptions must generate diagnostics
## Issue

If a programmer supplied assumption (assume!, etc.) is provably false, this should result in a diagnostic and the assumption should be ignored.

## Steps to Reproduce

See #389 

## Expected Behavior

A diagnostic message pointing out that the second assumption is provably false.

## Actual Results

The false assumption is asserted to the prover.

a-malyshev:
## Description

Fixes #393 

I'm not sure though that these changes are enough because initially, I thought that the solution should also address the problem described in #389, but according to the discussion, in the current implementation of the `result!`, it is not possible. 

If I'm missing something, I will be glad to hear feedback. 

## Type of change

## How Has This Been Tested?

./validate.sh
also, added an example to checker/run-pass tests

### DIFF
--- a/checker/src/call_visitor.rs
+++ b/checker/src/call_visitor.rs
@@ -559,16 +559,23 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
                     return;
                 }
 
-                // If the condition is always true, this assumption is redundant
-                if cond_as_bool.unwrap_or(false) {
-                    let span = self.block_visitor.bv.current_span;
-                    let warning =
-                        self.block_visitor.bv.cv.session.struct_span_warn(
-                            span,
-                            "assumption is provably true and can be deleted",
-                        );
-                    self.block_visitor.bv.emit_diagnostic(warning);
-                }
+                // If the condition is always true, this assumption is redundant. If false, the
+                // assumption is ignored. Otherwise, no diagnostics are emitted.
+                let message = if cond_as_bool == Some(true) {
+                    "assumption is provably true and can be deleted"
+                } else if cond_as_bool == Some(false) {
+                    "assumption is provably false and it will be ignored in the assertion"
+                } else {
+                    return;
+                };
+                let span = self.block_visitor.bv.current_span;
+                let warning = self
+                    .block_visitor
+                    .bv
+                    .cv
+                    .session
+                    .struct_span_warn(span, message);
+                self.block_visitor.bv.emit_diagnostic(warning);
             }
             KnownNames::MiraiPostcondition => {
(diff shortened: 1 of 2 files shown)


checker/tests/run-pass/assume_provably_false.rs
```diff
@@ -0,0 +1,5 @@
+pub fn main() {
+    let a = 5;
+    assume!(a < 5); //this assumption is provably false and it will be ignored
```
hermanventer, line 3: To make the test case real, it should be `//~ this assumption is provably false and it will be ignored`

It also seems that no message is generated, because the test case passes and because the expected error should be `assumption is provably false and it will be ignored in the assertion` according to the code.

checker/src/call_visitor.rs
```diff
@@ -559,16 +559,23 @@ impl<'call, 'block, 'analysis, 'compilation, 'tcx, E>
...
-                            "assumption is provably true and can be deleted",
-                        );
-                    self.block_visitor.bv.emit_diagnostic(warning);
-                }
+                // If the condition is always true, this assumption is redundant. If false, the
+                // assumption is ignored. Otherwise, no diagnostics are emitted.
+                let message = if cond_as_bool == Some(true) {
+                    "assumption is provably true and can be deleted"
+                } else if cond_as_bool == Some(false) {
+                    "assumption is provably false and it will be ignored in the assertion"
```
hermanventer, line 567: Your test case never gets here. That should be very interesting. I suspect that check_condition_value_and_reachability does the wrong thing because of the false assumption. It should be interesting to look into that.

checker/tests/run-pass/assume_provably_false.rs
a-malyshev: I misunderstood how tests are executed. Thanks for pointing this out.
Now, I see how it works. 
Fixed it.

hermanventer:
Does this not produce a warning for the second assumption in #389?

a-malyshev:
No, it doesn't. The warning for the second assumption is no produced because the result of `check_condition_value_and_reachability` is `None`. That's where I get confused. Then I read the [comment](https://github.com/facebookexperimental/MIRAI/issues/389#issuecomment-574314793) in the discussion of #393 and came to the conclusion that the current design does not allow this. But look like I misunderstood. 

I will take a closer look at how `None` is produced. I don't see why it is happening now.

hermanventer:
None means that the check was inconclusive, which could indicate a time-out or even that both p and ~p are satisfiable. The latter eventuality may be of interest. By all means look into that. But this PR may as well go in as is for now.

PR #609, opened 2020-08-11
Protect pre-state unknown values  from being refined using the post-state.

hermanventer:
# Description

Post conditions may relate values from the pre-state with values from the post state of a call. When a pre-state value is not known for some path whereas the post-state has a value for that path, it is important that value refinement not refine the pre-state derived value with a post-state. To stop this from happening we introduce Expression::RefinedVariable to distinguish values that have been determined to be undefined in a particular state and that should not be refined (by looking up their paths) in a later (i.e. post) state.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra


checker/src/abstract_value.rs
```diff
@@ -1197,7 +1209,7 @@ impl AbstractValueTrait for Rc<AbstractValue> {
             Expression::TaggedExpression { operand, tag } => {
                 operand.try_to_retype_as(target_type).add_tag(*tag)
             }
-            Expression::Variable { path, .. } => {
+            Expression::Variable { path, .. } | Expression::RefinedVariable { path, .. } => {
                 AbstractValue::make_typed_unknown(target_type.clone(), path.clone())
```
stonebuddha, line 1213: For RefinedVariable, should this be a call to make_typed_refined_unknown?

checker/src/expression.rs
```diff
@@ -669,6 +680,7 @@ impl Expression {
                 operand.expression.contains_local_variable()
             }
             Expression::Reference(path) => path.contains_local_variable(),
+            Expression::RefinedVariable { .. } => false,
```
stonebuddha, line 683: Why is this cases treated differently from Variable?

stonebuddha:
https://github.com/facebookexperimental/MIRAI/blob/56719f2b662bdc8b0fcaa62f3b7244d0d018c9bd/checker/src/abstract_value.rs#L2172-L2181
Do we need to also consider RefinedVariable here?

In general, I guess I'm a bit confused about when a Variable and a RefinedVariable should be treated differently. Could please add some comments about this?

checker/src/expression.rs
hermanventer: Only paths rooted in parameters can get wrapped up with RefinedVariable.

checker/src/abstract_value.rs
hermanventer: Ah yes, good catch.

checker/src/expression.rs
stonebuddha: https://github.com/facebookexperimental/MIRAI/blob/56719f2b662bdc8b0fcaa62f3b7244d0d018c9bd/checker/src/abstract_value.rs#L2881-L2884
I was assuming that but I got less confident when I read the code above.

BTW It would be nicer if this invariant for RefinedVariable is documented.

hermanventer:
Fair enough, I'll rename RefinedVariable to RefinedParameterCopy.

PR #615, opened 2020-08-13
Fix SMT encoding for variables

stonebuddha:
## Description

Variables need to be encoded as their original copies in SMT constraints. For example, if a loop's entry condition is obtained as `local == 0`, but the loop body modifies `local` and later `local` gets updated to `widen(.. at local)`, we want to distinguish the `local` in entry condition from the `local` in the widened value.

Some tests failed after the changes introduced by this commit. I investigated them and confirmed that they do need some extra mechanisms to work. See the todo comments in tests.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/z3_solver.rs
+++ b/checker/src/z3_solver.rs
@@ -588,7 +588,7 @@ impl Z3Solver {
 
     #[logfn_inputs(TRACE)]
     fn general_variable(&self, path: &Rc<Path>, var_type: &ExpressionType) -> z3_sys::Z3_ast {
-        let path_str = CString::new(format!("{:?}", path)).unwrap();
+        let path_str = CString::new(format!("{:?} original", path)).unwrap();
         unsafe {
             let path_symbol = z3_sys::Z3_mk_string_symbol(self.z3_context, path_str.into_raw());
             let sort = self.get_sort_for(var_type);
--- a/checker/tests/run-pass/smt_shift_left.rs
+++ b/checker/tests/run-pass/smt_shift_left.rs
@@ -12,7 +12,7 @@ pub fn read_uleb128_as_u32(bytes: [u8; 20]) -> u32 {
     let mut value: u32 = 0;
     let mut shift: u32 = 0;
     let mut cursor = 0;
-    while cursor < bytes.len() {
+    while cursor < bytes.len() && shift <= 28 {
         let byte = bytes[cursor];
         let val = byte & 0x7f;
         value |= (val as u32) << shift;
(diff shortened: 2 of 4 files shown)


checker/src/z3_solver.rs
```diff
@@ -588,7 +588,7 @@ impl Z3Solver {
 
     #[logfn_inputs(TRACE)]
     fn general_variable(&self, path: &Rc<Path>, var_type: &ExpressionType) -> z3_sys::Z3_ast {
-        let path_str = CString::new(format!("{:?}", path)).unwrap();
+        let path_str = CString::new(format!("{:?} original", path)).unwrap();
```
hermanventer, line 591: It is certainly very interesting that just adding " original" to the path will change anything but I'm not sure that is good thing. I want to understand this a bit better before just letting this in. Can you expand a bit on how you came across this problem and how this fixes it?

stonebuddha: https://github.com/facebookexperimental/MIRAI/blob/b1c75602295a68127c93c30fee402e5c3c2fd258/checker/src/abstract_value.rs#L500-L520

When trying to extract entry conditions of a loop, MIRAI uses `add_equalities_for_widened_vars` to obtain something like `local == 0`, if `local` is a variable initialized with 0. But in the loop body `local` might be updated to `local + 1`, and further `widen(.. at local)`. The SMT encoding for `widen(.. at local)` was just `local`, but I think it should be encoded differently from the variable `local` in the entry condition.

stonebuddha: https://github.com/facebookexperimental/MIRAI/blob/b1c75602295a68127c93c30fee402e5c3c2fd258/checker/src/fixed_point_visitor.rs#L251-L259

I guess the underlying issue is that entry conditions for a loop are not really entry conditions for the blocks in the loop. So currently if an entry condition references a variable `local`, the variable stands for its value in the entry environment.

hermanventer: Some of this is starting to come back to me. The use of the path to ensure that Z3 uses the same constant for both the initial and widened values is what I intended. The basic idea is that both loop guards and widened variables provide constraints on the otherwise unknown value of a loop variable. These constraints need to refer to the same constant, hence the use of the path.

The constraint that is obtained from the widened variable depends on the fragile heuristics for deriving an interval from a widened join expression. Hence this is not working in any non-trivial loop, particularly when there is more than one loop variable.

Can you provide the example that led you to look at this code?

stonebuddha: > The basic idea is that both loop guards and widened variables provide constraints on the otherwise unknown value of a loop variable.

The problem is that the entry condition of block is not really the loop guard; the usage of `add_equalities_for_widened_vars` adds constraints like `i == the initial value of i`, which is not true.

hermanventer: There is definitely a problem here that needs fixing. I just don't think the fix in the PR addresses the underlying problem.

stonebuddha: I see. I'll find out some better way to fix this.

stonebuddha:
SMT encoding for variables does not address the underlying issue. See #617 for an alternative fix.

PR #617, opened 2020-08-14
Fix entry condition construction for loops

stonebuddha:
## Description

Entry condition construction for loops was problematic because loop-back path conditions were not considered, and as a consequence, something like `loop_iterator == 0` may appear as an entry condition even if the loop body mutates `loop_iterator`.

Fixing this issue, in general, needs a mechanism for loop-invariant generation. This commit only implements a simple routine to identify paths that are not mutated by a loop and use them as the loop's invariants.

Some tests failed after the changes introduced by this commit. I investigated them and confirmed that they do need some extra mechanisms to work. See the todo/fixme comments in tests.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/environment.rs
+++ b/checker/src/environment.rs
@@ -200,11 +200,7 @@ impl Environment {
     #[logfn_inputs(TRACE)]
     pub fn widen(&self, other: Environment) -> Environment {
         self.clone()
-            .join_or_widen(other, |x, y, p| match (&x.expression, &y.expression) {
-                (Expression::Widen { .. }, _) => x.clone(),
-                (_, Expression::Widen { .. }) => y.clone(),
-                _ => x.clone().join(y.clone(), p).widen(p),
-            })
+            .join_or_widen(other, |x, y, p| x.join(y.clone(), p).widen(p))
     }
 
     /// Returns an environment with a path for every entry in self and other and an associated
--- a/checker/src/fixed_point_visitor.rs
+++ b/checker/src/fixed_point_visitor.rs
@@ -242,6 +242,27 @@ impl<'fixed, 'analysis, 'compilation, 'tcx, E>
             .0;
 
         let at_loop_anchor = self.loop_anchors.contains(&bb);
+
+        // Identifies loop invariants. Currently we only extract invariant paths, i.e., paths that are not
+        // mutated by the loop. todo: support more kinds of invariants.
+        let mut joined_state_without_loop_variants = joined_state.clone();
+        if at_loop_anchor {
+            self.bv.mir.predecessors()[bb].iter().for_each(|pred_bb| {
+                if self.dominators.is_dominated_by(*pred_bb, bb) {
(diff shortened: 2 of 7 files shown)


checker/src/environment.rs
```diff
@@ -200,11 +200,7 @@ impl Environment {
     #[logfn_inputs(TRACE)]
     pub fn widen(&self, other: Environment) -> Environment {
         self.clone()
-            .join_or_widen(other, |x, y, p| match (&x.expression, &y.expression) {
-                (Expression::Widen { .. }, _) => x.clone(),
-                (_, Expression::Widen { .. }) => y.clone(),
-                _ => x.clone().join(y.clone(), p).widen(p),
-            })
+            .join_or_widen(other, |x, y, p| x.clone().join(y.clone(), p).widen(p))
```
hermanventer, line 203: You don't have to clone x.

checker/src/fixed_point_visitor.rs
```diff
@@ -242,6 +242,27 @@ impl<'fixed, 'analysis, 'compilation, 'tcx, E>
...
+
+        // Identifies loop invariants. Currently we only extract invariant paths, i.e., paths that are not
+        // mutated by the loop. todo: support more kinds of invariants.
+        let mut joined_state_without_loop_variants = joined_state.clone();
+        if at_loop_anchor {
+            self.bv.mir.predecessors()[bb].iter().for_each(|pred_bb| {
+                if self.dominators.is_dominated_by(*pred_bb, bb) {
+                    let pred_out_state = &self.out_state[pred_bb];
+                    for (path, val) in pred_out_state.value_map.iter() {
+                        if let Some(widened_val) = joined_state.value_at(path) {
```
hermanventer, line 254: This value may not be widened, so the name is not that helpful. I had to read this a second time because of it.

checker/tests/run-pass/smt_shift_left.rs
```diff
@@ -12,17 +12,18 @@ pub fn read_uleb128_as_u32(bytes: [u8; 20]) -> u32 {
...
         let val = byte & 0x7f;
         value |= (val as u32) << shift;
         if val == byte {
             return value;
         }
         shift += 7;
-        if shift > 28 {
-            break;
-        }
+        // todo: this guard is not treated as the loop's entry condition because it is dominated by the loop anchor
```
hermanventer, line 23: I think what needs to happen here is that shift must get narrowed in the state that feeds back into the loop anchor.

checker/src/fixed_point_visitor.rs
stonebuddha: Renamed to `joined_val`.

checker/tests/run-pass/smt_shift_left.rs
stonebuddha: Updated the todo comment.

PR #619, opened 2020-08-18
Deal with non-standard enum encoding

stonebuddha:
## Description

Fieldless enums can have discriminants with explicitly defined tag values. For example, `std::cmp::Ordering` is defined as `{ Less = -1, Equal = 0, Greater = 1 }`. This commit implements a routine to extract consistent discriminant tags for fieldless enums.

Some enums have non-standard encoding. For example, a value of `Option<char>` is encoded as a 32-bit integer, where `Some(c)` is represented by the value of `c`, and `None` is represented by `0x110000`. This commit also takes care of such special encoding of enums.

This commit also fixes a few issues:
- `switchInt` instructions can have multiple branches with the same target.
- The treatment of `TOP` should be slightly different from `BOTTOM`. Currently, when the size of an expression becomes too large, it is then abstracted by `TOP`, which really means an existing but unknown value.

## Type of change

## How Has This Been Tested?

./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/block_visitor.rs
+++ b/checker/src/block_visitor.rs
@@ -25,10 +25,12 @@ use log_derive::*;
 use mirai_annotations::*;
 use rustc_hir::def_id::DefId;
 use rustc_middle::mir;
-use rustc_middle::mir::interpret::{ConstValue, Scalar};
+use rustc_middle::mir::interpret::{sign_extend, truncate, ConstValue, Scalar};
 use rustc_middle::ty::adjustment::PointerCast;
+use rustc_middle::ty::layout::PrimitiveExt;
 use rustc_middle::ty::subst::SubstsRef;
 use rustc_middle::ty::{Const, ParamConst, Ty, TyKind, UserTypeAnnotationIndex};
+use rustc_target::abi::{TagEncoding, VariantIdx, Variants};
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::convert::TryFrom;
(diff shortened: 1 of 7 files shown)


checker/src/block_visitor.rs
```diff
@@ -2208,15 +2223,38 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                        8 => u128::from(variant_tag as u64),
+                        _ => variant_tag,
+                    };
+                    if variant_data == *data {
+                        Some(variant_tag)
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or(*data); // todo: Deal with optimized layouts for enum types like Option<char>.
```
hermanventer, line 2251: I have a hard time understanding this code. It seems that the point here is to do sign extension for negative numbers. It boggles my mind that it has to be this complicated. Why is variant_discr.val not sufficient in all cases? Some comments with illustrative examples will be really helpful.

stonebuddha: Added some comments on the complication.

hermanventer: Wouldn't tcx.sign_extend be useful to simplify the code?

stonebuddha: I checked the source code of rustc and found out that we need to `sign_extend` and then `truncate` here.

Instead of just adding the workaround, I figured out that it would better to just implement the logic to deal with enum encoding.

checker/src/block_visitor.rs
```diff
@@ -2199,35 +2216,200 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                                    dataful_variant
+                                };
+                                discr_bits = truncate(variant.as_u32() as u128, discr_layout.size);
+                                discr_index = variant;
+                            }
+                        }
+                    }
+                };
+
+                debug!(
```
hermanventer, line 2344: trace!

checker/src/block_visitor.rs
```diff
@@ -2199,35 +2216,200 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+
+                    if discr_dataful {
+                        // Obtains the name of this variant.
+                        let name_str = {
+                            use std::ops::Deref;
+                            let enum_def = ty.ty_adt_def().unwrap();
+                            let variant_def = &enum_def.variants[discr_index];
+                            String::from(variant_def.ident.name.as_str().deref())
+                        };
+                        debug!("discr name {:?}", name_str);
```
hermanventer, line 2371: trace!

checker/tests/run-pass/iterator.rs
```diff
@@ -0,0 +1,77 @@
...
+//         end: 10usize,
+//     };
+
+//     while let Some(_) = it.next() {
+//         verify!(it.start <= 10);
+//     }
+//     verify!(it.start >= 10);
+// }
+
+// todo: sizes of expressions are too large. Simplify them on the fly.
```
hermanventer, line 26: Perhaps worth checking if this is still needed.

stonebuddha: test1 and test2 are commented out because they somehow caused CI test failures. On my local machine, MIRAI can pass these tests, and the issue that expression sizes are too large still exists.

PR #621, opened 2020-08-18
Fix nested loop fix point computation

hermanventer:
## Description

When checking if a nested loop body has reached a fixed point, use the out_state of the last block of the nested loop body. More importantly, make a copy of out_state before processing the current block/body because doing so will (nowadays) update out_state. This had the perverse effect of always terminating (non nested loop) fixed point iterations after 4 iterations. Fixing this uncovered diverging behavior with widening when widened values are updated via refinement of summaries.

The non convergence of widened values obtained from summaries has been fixed by not transferring summary side effects when the side effect will replace one widened value with another.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/src/fixed_point_visitor.rs
+++ b/checker/src/fixed_point_visitor.rs
@@ -88,7 +88,7 @@ impl<'fixed, 'analysis, 'compilation, 'tcx, E>
                         .expect("incorrectly initialized out_state")
                         .exit_conditions = self.bv.current_environment.exit_conditions.clone();
                 } else {
-                    self.compute_fixed_point(bb)
+                    self.compute_fixed_point(bb);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


checker/src/fixed_point_visitor.rs
```diff
@@ -146,38 +149,46 @@ impl<'fixed, 'analysis, 'compilation, 'tcx, E>
...
 
                 // Check for a fixed point.
-                if !self.bv.current_environment.subset(&self.out_state[&bb]) {
+                if !self.out_state[&last_block].subset(&old_state[&last_block]) {
                     // There is some path for which self.bv.current_environment.value_at(path) includes
-                    // a value this is not present in self.out_state[bb].value_at(path), so any block
+                    // a value this is not present in self.out_state[last_block].value_at(path), so any block
                     // that used self.out_state[bb] as part of its input state now needs to get reanalyzed.
                     changed = true;
                 }
```
stonebuddha, line 188: Could the last block be an arbitrary block in the loop body? Ideally, I think we need to check every block that has an edge that escapes the loop.

hermanventer: The loop body blocks are topologically sorted, so the last block is not entirely arbitrary. I think you are right to wonder what happens when there are multiple candidates for "last block" status. The block we end up choosing may well reach a fixed point before some of the other candidates. I don't think that is going to cause soundness issues, but it could decrease precision. Probably worth thinking some more and creating an issue, but I don't think I'm going to get around to deep thinking about it for quite some time to come.

PR #626, opened 2020-08-25
New nightly

hermanventer:
## Description

Lots of changes needed to deal with new patterns of MIR code generation and having to deal with encoded enum discriminators

## Type of change

## How Has This Been Tested?
./validate.sh (can't run MIRAI on itself right now)
Not working for libra_crypto either


checker/tests/run-pass/constant_time_taint.rs
```diff
@@ -39,7 +39,7 @@ pub fn test2(secret: i32) {
 pub fn test3(secret: i32) {
     precondition!(has_tag!(&secret, ConstantTimeTaint));
     let _ = if secret | 99991 < 5 { true } else { false };
-    //~ the branch condition has a ConstantTimeTaintKind tag
+    // ~ the branch condition has a ConstantTimeTaintKind tag
```
hermanventer, line 42: Di, could you have a look at why this has stopped working?

checker/src/constant_domain.rs
```diff
@@ -726,6 +726,12 @@ impl ConstantDomain {
             (ConstantDomain::F64(val1), ConstantDomain::F64(val2)) => {
                 f64::from_bits(*val1) != f64::from_bits(*val2)
             }
+            (ConstantDomain::I128(val1), ConstantDomain::U128(val2)) => {
+                *val1 >= 0 && ((*val1 as u128) != *val2)
+            }
+            (ConstantDomain::U128(val1), ConstantDomain::I128(val2)) => {
+                *val2 >= 0 && ((*val2 as u128) != *val1)
+            }
```
stonebuddha, line 734: Should the comparison be `*val1 < 0 || ((*val1 as u128) != *val2)`?

hermanventer: That would make a lot more sense...

checker/src/expression.rs
```diff
@@ -1121,51 +1121,39 @@ impl ExpressionType {
...
     /// Returns true if this type is not a primitive type. References are not regarded as
     /// primitives for this purpose.
     #[logfn_inputs(TRACE)]
     pub fn is_primitive(&self) -> bool {
         use self::ExpressionType::*;
-        match self {
-            NonPrimitive | ThinPointer => false, //todo: raw pointer should be primitive
-            _ => true,
-        }
+        !matches!(self, NonPrimitive | ThinPointer)
```
stonebuddha, line 1142: The todo comment went away. Is it not an issue any more?

checker/tests/run-pass/iterator.rs
```diff
@@ -71,7 +71,8 @@ pub fn test3() {
     while let Some(_) = it.next() {
         verify!(it.start <= 10);
     }
-    verify!(it.is_empty());
+    //todo: fix this
+    verify!(it.is_empty()); //~ this is unreachable, mark it as such by using the verify_unreachable! macro
```
stonebuddha, line 75: This seems to be fixed.

checker/src/expression.rs
hermanventer: It seems a bit stale.

checker/tests/run-pass/iterator.rs
hermanventer: It succeeds sometimes locally, but fails consistently in CI. I need to spend more time on this, but I wan't you to have a look while you are still around.

checker/tests/run-pass/constant_time_taint.rs
stonebuddha: The code generated by Rust is actually constant-time; it optimizes the conditional expression to just `secret | 99991 < 5`.
To get the error message back, we can change it to return non-Boolean values, such as:
``` rust
let _ = if secret | 99991 < 5 { 32767 } else { 10003 };
```

checker/tests/run-pass/iterator.rs
hermanventer: And now CI is flaky too.

stonebuddha: This reminds me of the fact that test1 and test2 in this file (commented out) work well locally but also fail in CI.

hermanventer: It must be a time-out that only happens on smaller CI machines.

stonebuddha: I investigated this case and there are some invalid loop invariants such as `it.start == 0`. Need some further investigation.

stonebuddha: https://github.com/facebookexperimental/MIRAI/blob/c705bec00f089494d9d3597439be884be5e0b06f/checker/src/fixed_point_visitor.rs#L266-L270
A possible fix for this is to update the condition above to
``` rust
if joined_val.is_widened() || !val.eq(joined_val) {
    ...
}
```
because widened values should not be consider to be equal to itself across loop iterations. This would lose precision because some widened values may come from previous loops, rather than the current loop.

checker/tests/run-pass/iterator.rs
```diff
@@ -11,19 +11,18 @@
...
-//     verify!(it.start >= 10);
-// }
+    while let Some(_) = it.next() {
+        verify!(it.start <= 10);
+    }
+    verify!(it.start >= 10);
+}
 
-// todo: sizes of expressions are too large. Simplify them on the fly.
 // pub fn test2() {
```
stonebuddha, line 26: I find some issues when debugging this test. I'll submit a PR when this PR is merged.

PR #628, opened 2020-08-26
Fix enum encoding to make range iterators work

stonebuddha:
## Description

This commit fixes issues on enum encoding to make range iterators work.

## Type of change

## How Has This Been Tested?

./validate.sh (can't run MIRAI on itself right now)
Not working for libra either

### DIFF
--- a/checker/src/block_visitor.rs
+++ b/checker/src/block_visitor.rs
@@ -152,7 +152,6 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
                 let param_env = self.bv.type_visitor.get_param_env();
                 if let Ok(ty_and_layout) = self.bv.tcx.layout_of(param_env.and(ty)) {
                     let discr_ty = ty_and_layout.ty.discriminant_ty(self.bv.tcx);
-                    let discr_layout = self.bv.tcx.layout_of(param_env.and(discr_ty)).unwrap();
                     let discr_bits = match ty_and_layout
                         .ty
                         .discriminant_for_variant(self.bv.tcx, variant_index)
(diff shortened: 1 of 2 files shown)


checker/src/block_visitor.rs
```diff
@@ -335,7 +335,20 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
...
+                .layout_of(param_env.and(switch_ty))
+                .expect("rustc should ensure this");
+            let val: Rc<AbstractValue> = if switch_ty.is_signed() {
+                Rc::new(
+                    ConstantDomain::I128(sign_extend(values[i], switch_ty_and_layout.size) as i128)
+                        .into(),
+                )
+            } else {
+                Rc::new(ConstantDomain::U128(values[i]).into())
+            };
```
hermanventer, line 351: This code sequence is starting to look familiar. Perhaps factor it out and re-use it?

checker/src/path.rs
```diff
@@ -744,6 +744,18 @@ impl PathRefinement for Rc<Path> {
                     qualifier.refine_parameters(arguments, result, pre_environment, fresh);
                 let refined_selector =
                     selector.refine_parameters(arguments, result, pre_environment, fresh);
+                // todo: more path canonicalization?
```
hermanventer, line 747: Since refine_parameters is always followed by refine_paths, it seemed reasonable to make refine_parameters as special purpose as possible and not do anything that refine_paths does. Just looking at this in isolation it not quite clear to me that this should not instead happen in refine_paths. I kind of expect that to already be the case, so things might be a bit more complicated than they seem. For example, *&path and path are not the same thing if path refers to a structure, because the * effectively copies the structure in that case.

stonebuddha: I see. I was looking at the intermediate values between `refine_parameters` and `refine_paths` and got confused. Reverted these changes.

checker/src/block_visitor.rs
```diff
@@ -735,6 +729,27 @@ impl<'block, 'analysis, 'compilation, 'tcx, E>
         self.bv.get_u128_const_val(val)
     }
 
+    pub fn get_int_const_val(&mut self, mut val: u128, ty: Ty<'tcx>) -> Rc<AbstractValue> {
```
hermanventer, line 732: A comment about the function will be most welcome.

stonebuddha: Added a method comment.

PR #635, opened 2020-09-02
Correct refinement of *&p to p, if p is an array

hermanventer:
## Description

This PR started out as a fix to lookup_path_and_refine_result, which incorrectly returned BOTTOM for local variable paths that have not been initialized. (That seemed reasonable at the time it was written, but meanwhile the convention has been established that the absence of an entry in the environment means "unknown value of a given type". The Rust compiler makes sure that no access is allowed on uninitialized paths.)

Meanwhile, encountering BOTTOM is no longer a trigger for a diagnostic, so the erroneous production of BOTTOM causes some errors to go unreported, which meant that some test cases succeeded when they should have failed.

The bulk of this PR deals with failures that result from refine_paths canonicalizing paths of the form *&p into just p. Most of the time this is exactly right, but if p is the location of an array and &p is the result of array::as_ptr(), then *&p is an alias for p[0]. 

Fixing this corner case uncovered a few others as well and turned a test case into something that needed a todo to become a done.

## Type of change

## How Has This Been Tested?
./validate.sh
ran MIRAI over Libra

### DIFF
--- a/checker/tests/run-pass/factorial.rs
+++ b/checker/tests/run-pass/factorial.rs
@@ -8,19 +8,19 @@
 
 // MIRAI_FLAGS --diag=strict
 
-#[macro_use]
-extern crate mirai_annotations;
-
-fn fact(n: u8) -> u128 {
-    if n == 0 {
-        1
-    } else {
-        let n1fac = fact(n - 1);
-        verify!(n1fac <= std::u128::MAX / (n as u128)); //~ possible false verification condition
-        (n as u128) * n1fac
-    }
-}
+// #[macro_use]
+// extern crate mirai_annotations;
+//
+// fn fact(n: u8) -> u128 {
+//     if n == 0 {
+//         1
+//     } else {
+//         let n1fac = fact(n - 1);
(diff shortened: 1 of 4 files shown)


hermanventer:
I need to think some more about this.

checker/tests/run-pass/factorial.rs
```diff
@@ -8,19 +8,19 @@
...
-fn fact(n: u8) -> u128 {
-    if n == 0 {
-        1
-    } else {
-        let n1fac = fact(n - 1);
-        verify!(n1fac <= std::u128::MAX / (n as u128)); //~ possible false verification condition
-        (n as u128) * n1fac
-    }
-}
+// #[macro_use]
```
wrwg, line 11: Left over? Or if this should stay commented out, perhaps add a TODO or note why?

wrwg (approved):

checker/tests/run-pass/factorial.rs
hermanventer: There is a new failure, which is probably expected, but there also seems to be a failure in the code for matching actual with expected output. I'll get to it one way or another.

PR #754, opened 2021-02-09
Adapt to nightly 2021-02-12

hermanventer:
## Description

This is branch that illustrates a problem with write_mir_pretty in recent nightly builds. See https://github.com/rust-lang/rust/issues/81200.

To reproduce the issue, clone MIRAI from github. Modify rust-toolchain to match the version you are using to test the fix. Then run cargo test.

Without a fix, you should see a bunch of failures such as:

thread 'rustc' panicked at 'DefId::expect_local: `DefId(5:3241 ~ alloc[4acc]::collections::vec_deque::{impl#4}::new)` isn't local', /rustc/0fc6756b42e0556cc2e18079f5fc6b4d58f4e81a/compiler/rustc_span/src/def_id.rs:175:43
n


checker/src/body_visitor.rs
```diff
@@ -161,8 +163,8 @@ impl<'analysis, 'compilation, 'tcx, E> BodyVisitor<'analysis, 'compilation, 'tcx
         function_constant_args: &[(Rc<Path>, Rc<AbstractValue>)],
         parameter_types: &[Ty<'tcx>],
     ) -> Summary {
-        if cfg!(DEBUG) {
-            let mut stdout = std::io::stdout();
+        if !cfg!(DEBUG) {
```
hermanventer, line 166: If you remove the prefixed negation, all tests should pass. If you keep the negation and fix the write_mir_pretty issue, all tests should pass.

PR #967, opened 2021-08-30
Call graph fixes for missing no-arg edge and duplicated croot node

darioncassel:
## Description

This PR fixes two bugs:
1. If a static call was made without any arguments, an edge was *not* added to the call graph.
2. If a public function had a definition that MIRAI reached (during analysis) after MIRAI already reached a call of that function, a *duplicate* node was added to the graph.

The fixes in this PR ensure that:
1. An edge does get added.
2. The node gets updated rather than a duplicate node being added.

Additionally, this PR adds missing help text for one of the call graph configuration options.

## Type of change

## How Has This Been Tested?

Two test cases have been added that target these bugs. The test cases will fail unless the fixes are in place.

## Checklist:

- [x] Fork the repo and create your branch from `master`.

### DIFF
--- a/checker/src/call_graph.rs
+++ b/checker/src/call_graph.rs
@@ -358,8 +358,19 @@ impl CallGraph {
     /// Add a new crate root node to the call graph.
     pub fn add_croot(&mut self, defid: DefId) {
         let croot = CallGraphNode::new_croot(defid);
-        let node_id = self.graph.add_node(croot);
-        self.nodes.insert(defid, node_id);
+        match self.nodes.entry(defid) {
+            Entry::Occupied(node) => {
+                // Replace non-croot existing node
+                let node_id = node.get().to_owned();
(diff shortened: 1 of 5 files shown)


hermanventer (approved):
The changes seems correct with respect to the specification. The specification seems debatable. I.e. should a crate have a single root? If so, should that root be the last one for which add_croot has been called? In particular, if a crate specifies its entry function, shouldn't that function take precedence over others (all other things being equal)?

darioncassel:
This fix doesn't prevent a crate from having more than one root (nor did the state of the code before this fix). 

Rather the fix here is for this scenario: While analyzing the code we encounter a call. We don't (yet) know that the called function *is* a crate root so it gets added as a regular root. Later, when we find out that it is a crate root, the bug is that we were adding a new node for it rather than updating the existing node.

hermanventer:
Ah yes, got it now. Makes sense.

PR #1007, opened 2021-09-19
[callgraph] Improve derivation of type relations

darioncassel:
## Description

The call graph generator derives type relationships between call graph edge types; in particular, equality and membership relations. There were two issues with this process:

1. The heuristics used in this process were ad-hoc and thus resulted in some expected type relations not being derived. For example, the `EqType` relation captures equality of types modulo references and mutability. Thus, `EqType(&mut T, &T)` should hold, but this was not derived from the heuristics.
2. User-input type relations were added verbatim without attempting to derive any additional relations given the heuristics and edge types. Thus, relations that would intuitively have been expected to hold failed to be derived. For example, if a user input `Member(T1, T1a)` and heuristic-derived relation was `EqType(T1, T2)`, there was no derivation of `Member(T2, T1a)`.

This has been fixed by redesigning the type relation derivation process to 1) be more principled, and 2) work consistently over call graph types and input type relations. The new architecture is as follows:

1. Call graph edge types are added to a `type_map`
1. Input type relations are gathered into `type_relations`. Types from these relations are added to the `type_map`.
2. Types in the `type_map` are simplified using the heuristics.
3. `EqType` and `Member` relations are derived using the `type_map`.

## Type of change

## How Has This Been Tested?

A new test `tests/call_graph/type_relations.rs` has been added to test type relation derivation. It does also test input type relations.

## Checklist:

- [x] Fork the repo and create your branch from `main`.

### DIFF
--- a/checker/src/call_graph.rs
+++ b/checker/src/call_graph.rs
@@ -8,6 +8,7 @@ use petgraph::dot::{Config, Dot};
 use petgraph::graph::{DefaultIx, NodeIndex};
 use petgraph::visit::Bfs;
 use petgraph::{Direction, Graph};
+use regex::Regex;
 use rustc_hir::def_id::DefId;
 use serde::ser::{SerializeMap, Serializer};
 use serde::{Deserialize, Serialize};
@@ -251,6 +252,21 @@ impl TypeRelation {
     }
 }
 
+const COLLECTION_TYPES: &[&str] = &[
+    "std::slice::Iter",
+    "std::iter::Enumerate",
+    "std::iter::Map",
+    "std::collections::HashSet",
+    "std::collections::HashMap",
+    "std::vec::Vec",
+];
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum SimpleType {
+    Base(Box<str>),
+    Collection(Box<str>),
+}
+
 /// Types associated with call graph edges
 #[derive(Debug, Clone)]
 struct EdgeType {
@@ -258,63 +274,11 @@ struct EdgeType {
     id: TypeId,
     // Textual representation of the type
     name: Box<str>,
-    /// Derived type relations
-    type_relations: HashSet<TypeRelation>,
 }
 
 impl EdgeType {
     fn new(id: TypeId, name: Box<str>) -> EdgeType {
-        let type_relations = EdgeType::derive_type_relations(&name);
-        EdgeType {
-            id,
-            name,
-            type_relations,
-        }
-    }
-
-    /// Derive equality and membership relationships from the following
-    /// heuristics:
-    /// - Option<t> == t
-    /// - &t == t
-    /// - mut t == t
-    /// - [t] > t
-    fn derive_type_relations(name: &str) -> HashSet<TypeRelation> {
-        let mut relations = HashSet::<TypeRelation>::new();
-        let mut base_type = name.to_owned();
-        if base_type.contains("&std::option::Option<") {
-            let base_type_new = base_type
-                .replace("&std::option::Option<", "")
-                .replace(">", "");
-            relations.insert(TypeRelation::new_eq(
-                base_type.into_boxed_str(),
-                base_type_new.clone().into_boxed_str(),
-            ));
-            base_type = base_type_new;
-        }
-        if base_type.contains('[') && base_type.contains(']') {
-            let base_type_new = base_type.replace("[", "").replace("]", "");
-            relations.insert(TypeRelation::new_member(
-                base_type.into_boxed_str(),
-                base_type_new.clone().into_boxed_str(),
-            ));
-            base_type = base_type_new;
-        }
-        if base_type.contains("mut") {
-            let base_type_new = base_type.replace("mut ", "");
-            relations.insert(TypeRelation::new_eq(
-                base_type.into_boxed_str(),
-                base_type_new.clone().into_boxed_str(),
-            ));
-            base_type = base_type_new;
-        }
-        if base_type.contains('&') {
-            let base_type_new = base_type.replace("&", "");
-            relations.insert(TypeRelation::new_eq(
-                base_type.into_boxed_str(),
-                base_type_new.into_boxed_str(),
-            ));
-        }
-        relations
+        EdgeType { id, name }
     }
 }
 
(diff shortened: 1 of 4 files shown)


checker/src/call_graph.rs
```diff
@@ -720,13 +684,172 @@ impl CallGraph {
...
     fn gather_type_relations(
         &self,
-        index_to_type: &HashMap<TypeId, &EdgeType>,
+        type_map: &mut HashMap<TypeId, Box<str>>,
         type_relations_path: Option<&Path>,
     ) -> HashSet<TypeRelation> {
+        let mut type_to_index = HashMap::<Box<str>, TypeId>::new();
+        for (type_id, type_str) in type_map.iter() {
+            type_to_index.insert(type_str.to_owned(), *type_id);
+        }
```
hermanventer, line 847: You can build the HashMap using a more functional expression such as
```
    let type_to_index: HashMap::<Box<str>, TypeId> = 
         type_map.iter().map(|(id,bx)| (bx.to_owned(), *id)).collect();
```

checker/src/call_graph.rs
```diff
@@ -736,15 +859,27 @@ impl CallGraph {
...
+                    None => {
+                        max_id += 1;
+                        type_map.insert(max_id, relation.type1.to_owned());
+                    }
+                }
+                match type_to_index.get(&relation.type2) {
+                    Some(_) => {}
+                    None => {
+                        max_id += 1;
+                        type_map.insert(max_id + 2, relation.type2.to_owned());
```
hermanventer, line 876: I find myself wondering why you add 2 to max_id here.

checker/src/call_graph.rs
```diff
@@ -720,13 +684,172 @@ impl CallGraph {
...
+        type_map: &mut HashMap<TypeId, Box<str>>,
         type_relations_path: Option<&Path>,
     ) -> HashSet<TypeRelation> {
+        let mut type_to_index = HashMap::<Box<str>, TypeId>::new();
+        for (type_id, type_str) in type_map.iter() {
+            type_to_index.insert(type_str.to_owned(), *type_id);
+        }
         let mut type_relations = HashSet::<TypeRelation>::new();
-        // Insert input type relations from file
+        let mut max_id = type_map
```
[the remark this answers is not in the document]
darioncassel: Good idea

checker/src/call_graph.rs
darioncassel: Thanks for the suggestion, I'll rewrite that.

checker/src/call_graph.rs
darioncassel: That's a bug that was introduced through refactoring; thanks for pointing that out!

checker/src/call_graph.rs
```diff
@@ -720,13 +684,175 @@ impl CallGraph {
...
+        false
+    }
+
+    /// Derive all type relations
+    fn derive_all_relations(
+        &self,
+        type_map: &HashMap<TypeId, Box<str>>,
+        eq_map: &HashMap<Box<str>, Box<str>>,
+    ) -> HashSet<TypeRelation> {
+        let mut simple_types = Vec::<(Box<str>, SimpleType)>::new();
```
hermanventer, line 764: Since simple_types does not outlive type_map, you can just use &str rather than Box<str>

checker/src/call_graph.rs
```diff
@@ -720,13 +684,175 @@ impl CallGraph {
...
+        for (_, t) in type_map.iter() {
+            let simple_ts = self.simplify_type(t.as_ref());
+            for simple_t in simple_ts.iter() {
```
hermanventer, line 767: consider using into_iter here. That transfers ownership so you don't have to clone things.

checker/src/call_graph.rs
```diff
@@ -720,13 +684,175 @@ impl CallGraph {
...
+                simple_types.push((t.to_owned(), simple_t.to_owned()));
```
hermanventer, line 768: It seems like you are extending simple_types with a mapping of simple_ts. You can do that in a more functional way.

checker/src/call_graph.rs
```diff
@@ -720,13 +684,175 @@ impl CallGraph {
...
+            }
+        }
+        let mut relations = HashSet::<TypeRelation>::new();
+        // Initialize the set of pairs with the existing equality relations
+        let mut pairs = eq_map
```
hermanventer, line 773: I don't see the point of pairs.

checker/src/call_graph.rs
```diff
@@ -720,13 +684,175 @@ impl CallGraph {
...
+        for relation in input_relations.iter() {
+            match relation.kind {
+                TypeRelationKind::Eq => {
+                    eq_map.insert(relation.type1.to_owned(), relation.type2.to_owned());
+                    eq_map.insert(relation.type2.to_owned(), relation.type1.to_owned());
+                }
+                TypeRelationKind::Member => {}
+            }
+        }
+        eq_map
```
hermanventer, line 838: This looks like filter_map to me.

darioncassel: It doesn't quite work as a filter_map because we are inserting two values into the eq_map per one value in the type_relations. Thus, we'd need to run the filter_map twice, which seems less efficient to me.

checker/src/call_graph.rs
darioncassel: The pairs set is used to avoid redundant EqType(a,b) relations by keeping track of the pairs (a,b) and (b,a) that have been added so far. It is initialized with the existing EqType relations that come from user input.

hermanventer: Normalization and insert seem to do that already.

checker/src/call_graph.rs
hermanventer: Try flat_map

darioncassel: That worked, thanks!

checker/src/call_graph.rs
darioncassel: Unfortunately no, because the set of relations is initially empty; it doesn't contain the input type relations. However, I will refactor this so that the optimization you've suggested does work.

hermanventer (approved):

PR #1123, opened 2022-01-06
Fix building the checker with stable rust

dcsommer:
## Description

Stable rustc is capable of building the checker tool; no nightly compiler appears to be needed.

## Type of change

- [ x ] Code cleanup or just keeping up with the latest Rustc nightly

## How Has This Been Tested?

On Fedora laptop:
```
rustup update stable
CPATH=/usr/include/z3 cargo install  --path ./checker
CPATH=/usr/include/z3 cargo test  # this had errors on both nightly and stable 1.58
```

## Checklist:

- [ x ] Fork the repo and create your branch from `main`.
- [ x ] If you've added code that should be tested, add tests.
- [ x ] If you've changed APIs, update the documentation.
- [ x ] Ensure the test suite passes.
- [ x ] Make sure your code lints.
- [ x ] If you haven't already, complete your CLA here: <https://code.facebook.com/cla>

### DIFF
--- a/checker/src/utils.rs
+++ b/checker/src/utils.rs
@@ -307,7 +307,7 @@ fn append_mangled_type<'tcx>(str: &mut String, ty: Ty<'tcx>, tcx: TyCtxt<'tcx>) 
         }
         Param(param_ty) => {
             str.push_str("generic_par_");
-            str.push_str(param_ty.name.as_str());
+            str.push_str(&param_ty.name.as_str());
         }
         Projection(projection_ty) => {
             append_mangled_type(str, projection_ty.self_ty(), tcx);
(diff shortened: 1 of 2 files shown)


hermanventer:
MIRAI can be built with a rust compiler other than the one in toolchain, but it cannot run successfully unless it has been built with the nightly build. The reason for this is that MIRAI relies on unstable features of the rust compiler that is only exposed in the nightly build.

dcsommer:
Makes sense. Does this change make sense to merge? Or is the compile error for stable a "feature" to prevent confusion later when the tool doesn't work properly?

hermanventer:
The source code obviously has to match the Rust compiler version specified in the toolchain file. That may not be compatible with the current stable version of the Rust compiler, but not necessarily so.

I guess the right fix here is to add conditional code to the the main routine that will terminate with a helpful error message when it has been compiled with a rustc version other than what is in the toolchain file.

hermanventer:
Dealt with elsewhere.

PR #1158, opened 2022-06-21
Cargo mirai

hermanventer:
## Description

When using a recursive call to cargo to run a test target, include a `--test <name>` flag to prevent other test targets from getting built with rustc, which will cause the next recursive call (with the next test target) to become a no-op because rustc has already built it.

## Type of change

## How Has This Been Tested?
Ran `cargo mirai` over a project with more than one test target and observed that each test target in the project is built with MIRAI rather than rustc.


hermanventer:
No one at Meta seems interested to merge pull requests.

PR #1221, opened 2023-06-15
annotations: Distribute the license with the crate

dm0-:
## Description

This just adds `license-file` to the `annotations` crate pointed at the license in the parent directory.  Since rust-lang/cargo#7905, this will distribute the license text in the crate so it is available with the code fetched from crates.io.  (It could be added to the other `Cargo.toml`s, too, but `mirai-annotations` is the only one I see being published on crates.io from a quick search.)

## Type of change

## How Has This Been Tested?

Run `cargo package --list` to verify the `LICENSE` file will be added, and also just `cargo package` to inspect it.

## Checklist:

- [x] Fork the repo and create your branch from `main`.
- [ ] Ensure the test suite passes.


dm0-:
I've updated this to use a symlink to copy the license into the crate instead, since defining `license-file` semantically means it's a nonstandard license apparently.  If symlinks should be avoided, the file could be copied.

PR #1272, opened 2024-07-15
nightly-2023-12-30

hermanventer:
## Description

Update toolchain to nightly-2023-12-30 and deal with changes. Also update dependencies.

## Type of change

## How Has This Been Tested?
./validate.sh vcpkg


gal1ium:
Hi! I found that, in `standard_contracts/src/foreign_contracts.rs`, some std contracts are not updated according to the nightly-2023-12-30's std. For example, https://github.com/facebookexperimental/MIRAI/blob/41887ea41f5abbdc5006e6c297cfed2992d40c75/standard_contracts/src/foreign_contracts.rs#L125 's `cap` in nightly-2023-12-30 is an Enum `Cap`, not a `usize`, which leads to some wrong analysis results.
May I ask if the only way is to compare the current foreign_contracts with the std source one by one to fix such issues?

hermanventer:
Sadly, yes. At some point it would be great if we can build MIRAI in such a way that the full sources of the standard libraries are available to MIRAI at runtime. Perhaps this is feasible now, but when I had a lot more time to work on this, it was not possible.

